PR #11, opened 2021-11-29
WIP: Pod Market

BeanstalkFarms:


protocol/contracts/farm/facets/MarketplaceFacet/MarketplaceFacet.sol
```diff
@@ -0,0 +1,292 @@
...
+        emit BuyOfferCreated(s.buyOfferIndex, msg.sender, amount, pricePerPod, maxPlaceInLine);
+
+        s.buyOfferIndex = s.buyOfferIndex + 1;
+
+    }
+
+    function buyBeansAndListBuyOffer(uint232 maxPlaceInLine, uint24 pricePerPod, uint256 amountBeans, uint256 buyBeanAmount) public payable {
+
+        require(bean().balanceOf(msg.sender) >= amountBeans, "Marketplace: Not enough beans to submit buy offer.");
+        require(pricePerPod > 0, "Marketplace: Price must be greater than 0");
```
BeanstalkFarms, line 149: Should we be enforcing max here too?

protocol/contracts/farm/facets/MarketplaceFacet/MarketplaceFacet.sol
```diff
@@ -0,0 +1,292 @@
...
+        console.log('bean().balanceOf(msg.sender) after',bean().balanceOf(address(this)));
+
+    }
+
+
+    function sellToBuyOffer(uint256 plotIndex, uint24 buyOfferIndex, uint232 amount) public  {
+        
+        require(s.a[msg.sender].field.plots[plotIndex] >= 0, "Marketplace: Plot  not owned by user.");
+
+
```
BeanstalkFarms, line 176: Try to avoid ever having more. than 1 blank line in a row.

protocol/contracts/farm/facets/MarketplaceFacet/MarketplaceFacet.sol
```diff
@@ -0,0 +1,292 @@
...
+        _transferPlot(msg.sender, buyOffer.owner, plotIndex, amount);
+
+        emit BuyOfferFilled(buyOfferIndex, amount);
+
+    }
+
+    function cancelBuyOffer(uint24 buyOfferIndex) public  {
+        Storage.BuyOffer storage buyOffer = s.buyOffers[buyOfferIndex];
+        // TODO does this check equality correctly?
+        require(buyOffer.owner == msg.sender, "Field: Buy Offer not owned by user.");
```
BeanstalkFarms, line 227: Yes

protocol/contracts/farm/facets/MarketplaceFacet/MarketplaceFacet.sol
```diff
@@ -0,0 +1,292 @@
...
+            //     s.listedPlots[plotIndex.add(amount)] = listing;
+            //     s.listedPlots[plotIndex.add(amount)].amount = listingAmount - amount;
+            //     delete s.listedPlots[plotIndex];
+            // }
+            // emit ListingFilled(buyOffer.owner, msg.sender, plotIndex, listing.price, amount);
+
+
+        }
+
+        buyOffer.amount = buyOffer.amount - amount;
```
BeanstalkFarms, line 216: So we should discuss what the best thing to do here -> this line + line 186 is the same thing as using SafeMath. Throughout the project we have used SafeMath, but this is fine too.

protocol/contracts/farm/facets/MarketplaceFacet/MarketplaceFacet.sol
```diff
@@ -0,0 +1,292 @@
...
+
+        require(plotSize >= amount, "Marketplace: Plot not large enough.");
+
+        require(pricePerPod > 0, "Marketplace: Plot price must be non-zero.");
+        require(pricePerPod < 1000000, "Marketplace: Plot price must be less than 1.");
+
+        uint232 harvestable = uint232(s.f.harvestable);
+        require(harvestable <= index, "Marketplace: Plot cannot be harvestable.");
+        require(expiry <= index, "Marketplace: Plot must expire before harvestable.");
+        require(s.listedPlots[index].price == 0, "Marketplace: Plot already listed.");
```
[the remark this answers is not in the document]
adameldefrawy: Yeah, I think this is good idea to keep it in same function, so list plot will just cancel and update if already listed.thus, In accidental case of sending two list plot transactions the latter one goes through

PR #15, opened 2021-12-01
optimizing withdrawals for silos voted on

dantrinh:
removed unnecessary modifiers and added withdrawal ability on silos that one has voted on, added another test file


leonardo-fibonacci (commented):
Proposers should not be able to withdraw more than the minimum stalk required in the BIP, which is .1 percent of the total roots

leonardo-fibonacci (commented):
Add a proposition variable: 
-like vote until... proposedUntil
-set when user proposed and in the decrement roots function  at the end...if proposedUntil is greater than the season, then after decrementing... must check (users roots >  totalroots/ 1000)

leonardo-fibonacci (commented):
Use BIPs.sol the canPropose function as reference

PR #81, opened 2022-08-16
BIP-X: Pod Marketplace Price Functions

maltseasy:


protocol/contracts/farm/facets/MarketplaceFacet/Dynamic.sol
```diff
@@ -136,6 +141,10 @@
         return PPoly32(ranges, values, bases, f.signs, f.mode);
     }
 
+    /**
+    * @notice Creates a 'zero set' of PPoly32 values.
+    * @dev To be used when hashing constant listings/orders to prevent hashing collisions
+    */
     function createZeros() internal pure returns (uint256[32] memory ranges, uint256[128] memory values, uint256[4] memory bases) {
```
beanjoyer, line 148: I don't actually believe we need this to prevent hashing collisions. Its something we can bring up to Halborn to be 100% confident but as long as what we are encoding/hashing doesn't have arguments that have variable LENGTHS  we are OK. Having a variable NUMBER of arguments is alright. Can cover this tmrw

protocol/contracts/farm/facets/MarketplaceFacet/Dynamic.sol
```diff
@@ -0,0 +1,241 @@
...
+            constants: [1, 1, 25, 25, 2, 1, 50, 125, 1, 0, 0, ... , 0]
+            (expanded) bases: [0, 0, 0, 2, 0, 0, 0, 4, 0, 0, 0, ... , 0]
+            (expanded) signs: [true, true, true, true, false, true, true, true, false, false, false, ... , false]
+        
+    */
+    
+    //
+    struct PiecewisePolynomial {
+        uint256[16] breakpoints; //The breakpoints for the piecewise. The last breakpoint in the array is the final interval for the polynomial, extending to infinity.
+        uint256[64] constants; // [c0_0,c1_0,c2_0,c3_0, ... , c0_31,c1_31,c2_31,c3_31] Contains the coefficients of the polynomial, concatenated according to their piecewise index. The first coefficient is the constant term, the second is the first term, etc.
```
beanjoyer, line 46: `constants` => `coefficients` or `constantCoefficients` ? coefficients should probably belong in this variable since thats what these are

protocol/contracts/farm/facets/MarketplaceFacet/Dynamic.sol
```diff
@@ -0,0 +1,241 @@
...
+            (expanded) bases: [0, 0, 0, 2, 0, 0, 0, 4, 0, 0, 0, ... , 0]
+            (expanded) signs: [true, true, true, true, false, true, true, true, false, false, false, ... , false]
+        
+    */
+    
+    //
+    struct PiecewisePolynomial {
+        uint256[16] breakpoints; //The breakpoints for the piecewise. The last breakpoint in the array is the final interval for the polynomial, extending to infinity.
+        uint256[64] constants; // [c0_0,c1_0,c2_0,c3_0, ... , c0_31,c1_31,c2_31,c3_31] Contains the coefficients of the polynomial, concatenated according to their piecewise index. The first coefficient is the constant term, the second is the first term, etc.  
+        uint256[2] packedBases; // [b0,b1,b2,b3] b0 contains the 8-bit bases for polynomials 0-7, b1 for polynomials 8-15
```
beanjoyer, line 47: these are not actually the bases, this is the exponent part of the number

protocol/contracts/farm/facets/MarketplaceFacet/Dynamic.sol
beanjoyer: also should mention (perhaps in a comment) that these coefficients are the significands of a floating point representation where they are divided by 10^(corresponding exponent)

protocol/contracts/farm/facets/MarketplaceFacet/Dynamic.sol
beanjoyer: base is 10

beanjoyer: image

PR #82, opened 2022-08-16
BIP-24: Combine Token Support

0xkokonut:


protocol/contracts/farm/AppStorage.sol
```diff
@@ -74,6 +74,7 @@ contract Account {
         mapping(address => mapping(uint32 => Deposit)) deposits; // A Farmer's Silo Deposits stored as a map from Token address to Season of Deposit to Deposit.
         mapping(address => mapping(uint32 => uint256)) withdrawals; // A Farmer's Withdrawals from the Silo stored as a map from Token address to Season the Withdrawal becomes Claimable to Withdrawn amount of Tokens.
         SeasonOfPlenty sop; // A Farmer's Season Of Plenty storage.
+        mapping(address => mapping(address => uint256)) depositAllowances; // Spender => Silo Token
```
silochad, line 77: I missed an asymmetry here in my initial review:

Deposits is a mapping of: `Token Address => Season of Deposit => Amount`
Allowances is a mapping of: `Spender => Token Address`

In general I'd advocate for as much consistency in structure here as we can muster unless there are technical reasons to structure things differently.

However, given that this BIP is ~2 days from closing and this will require inverting the parameter order for all usages of depositAllowances, I think we let it go and make a note in future documentation.

protocol/contracts/farm/facets/SiloFacet/SiloFacet.sol
```diff
@@ -82,25 +82,66 @@ contract SiloFacet is TokenSilo {
      */
 
     function transferDeposit(
+        address sender,
```
silochad, line 85: @BeanstalkFarms/frontend: Requires ABI update and small tweak to the Transfer form.

protocol/contracts/farm/facets/SiloFacet/SiloFacet.sol
```diff
@@ -82,25 +82,66 @@ contract SiloFacet is TokenSilo {
...
         address recipient,
         address token,
         uint32[] calldata seasons,
         uint256[] calldata amounts
-    ) external payable nonReentrant updateSilo {
+    ) external payable nonReentrant returns (uint256[] memory bdvs) {
+        require(amounts.length > 0, "Silo: amounts array is empty");
+        for (uint256 i = 0; i < amounts.length; i++) {
+            require(amounts[i] > 0, "Silo: amount in array is 0");
+            if (sender != msg.sender) {
```
silochad, line 110: Gas optimization: move this check outside the loop, since `sender` and `msg.sender` don't change across iterations.

protocol/contracts/farm/facets/SiloFacet/TokenSilo.sol
```diff
@@ -355,6 +366,33 @@ contract TokenSilo is Silo {
...
+    function _spendDepositAllowance(
+        address owner,
+        address spender,
+        address token,
+        uint256 amount
+    ) internal virtual {
+        uint256 currentAllowance = depositAllowance(owner, spender, token);
+        if (currentAllowance != type(uint256).max) {
+            require(currentAllowance >= amount, "Silo: insufficient allowance");
+            _approveDeposit(owner, spender, token, currentAllowance - amount);
```
silochad, line 381: Note that this will re-emit the DepositApproval event after every authorized transfer - this isn't required for tokens by the ERC20 standard, but seems to be common behavior.

protocol/contracts/libraries/Convert/LibConvert.sol
```diff
@@ -66,7 +70,10 @@ library LibConvert {
         /// urBEAN -> urBEAN:3CRV LP
         if (tokenIn == C.unripeBeanAddress() && tokenOut == C.unripeLPAddress())
             return LibUnripeConvert.beansToPeg();
-        
+
+        // Lambda -> Lambda
+        if (tokenIn == tokenOut) return type(uint256).max;
+
         require(false, "Convert: Tokens not supported");
```
silochad, line 77: Is there a reason why this is `require(false, ...)` instead of `revert(...)`?

PR #95, opened 2022-09-20
implement plantDelegated

0xm00neth:


protocol/contracts/farm/facets/SiloFacet/DelegateFacet.sol
```diff
@@ -0,0 +1,71 @@
...
+ * @author Publius
+ * @title DelegateFacet handles delegating authority for Beanstalk functions.
+ */
+contract DelegateFacet is Silo {
+    mapping(bytes4 => bool) public selectorEnabled;
+
+    bytes4 public constant PLANT_DELEGATED_SELECTOR =
+        bytes4(keccak256(bytes("plantDelegated(address)")));
+
+    constructor() {
```
publiuss, line 22: Diamond facets can't use constructors because they are implementation contracts. Instead, we use separate init contracts that get executed 1 time during the `diamondCut` function. Instead of having a constructor, can we add an `InitFunctionApproval` contract with an `init()` function.

protocol/contracts/farm/facets/SiloFacet/DelegateFacet.sol
```diff
@@ -0,0 +1,71 @@
...
+        address delegatee,
+        bytes32 approval
+    ) external {
+        require(
+            selectorEnabled[selector],
+            "DelegateFacet: invalid function selector"
+        );
+        LibDelegate.setAllowance(msg.sender, selector, delegatee, approval);
+    }
+
```
publiuss, line 43: Can we add a `delegateAllowance` view function that returns the current allowance?

protocol/contracts/farm/facets/SiloFacet/DelegateFacet.sol
```diff
@@ -0,0 +1,71 @@
...
+    /**
```
publiuss, line 44: Can we add a `permit` function that uses a permit to approve as well?

See here for example:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/draft-ERC20Permit.sol#L49

protocol/contracts/libraries/Silo/LibDelegate.sol
```diff
@@ -0,0 +1,56 @@
...
+    ) internal {
+        AppStorage storage s = LibAppStorage.diamondStorage();
+        s.a[account].functionApprovals[selector][delegatee] = approval;
+    }
+
+    function spendAllowance(
+        address account,
+        bytes4 selector,
+        address spender,
+        uint256 beans
```
publiuss, line 46: Instead of `beans` can we use the word `amount`? Not all delegate functions will be denominated in Beans.

Issue #120, opened 2022-10-13
RFC: Migrate Fertilizer uri to decentralized SVG


PatrickAlphaC:
Right now, as part of [fertilizer](0x402c84De2Ce49aF88f5e2eF3710ff89bFED36cB6) contract, the `uri` function returns a result from a centralized server.

Using the following [cast](https://book.getfoundry.sh/) command
```
cast call 0x402c84De2Ce49aF88f5e2eF3710ff89bFED36cB6 "uri" 0 --rpc-url $MAINNET_RPC_URL
```

We get the result:

```
https://fert.bean.money/0
```

Or a more interesting result if we chose an ID like [3452735](https://fert.bean.money/3452735):

```json
{
  "name": "Fertilizer - 3452735",
  "external_url": "https://fert.bean.money/3452735.html",
  "description": "A trusty constituent of any Farmer's toolbox, ERC-1155 FERT has been known to spur new growth on seemingly dead farms. Once purchased and deployed into fertile ground by Farmers, Fertilizer generates new Sprouts: future Beans yet to be repaid by Beanstalk in exchange for doing the work of Replanting the protocol.",
  "image": "https://fert.bean.money/fertilizer.png?1665687752145",
  "attributes": [
    { "trait_type": "Season", "value": 6086 },
    {
      "trait_type": "Humidity",
      "display_type": "boost_percentage",
      "value": 244
    },
    {
      "trait_type": "BPF Remaining",
      "display_type": "boost_number",
      "value": 3.32
    },
    { "trait_type": "Updated At", "display_type": "date", "value": 1665687752 }
  ]
}
```

And each `tokenID` has a similar resulting centralized URL. 

I'd like to propose an SVG pattern instead - since as we know, maintaining a website is both costly and a very very minor centralized attack vector (not worth bug bountying). All the fertilizers could turn into broken jpegs :( 

In the `Internalizer.sol` contract is where the `uri` function lives, which gives us what the images of the fertilizers look like. It's currently defined as such:

```javascript
    function uri(uint256 _id)
        external
        view
        virtual
        override
        returns (string memory)
    {
        return string(abi.encodePacked(_uri, StringsUpgradeable.toString(_id)));
    }
```

Where `_uri` is hard coded to `https://fert.bean.money/`.  

Instead, we could use base64 encoding (one already created by a member of the [Loopring team](https://www.npmjs.com/package/base64-sol)) do something as such:

```javascript
import "base64-sol/base64.sol";
.
[...]

PR #122, opened 2022-10-15
⛓ on-chain metadata for fertilizers

PatrickAlphaC:
I left a `Note:` in there as well:

1. For `formatBpfRemaining`, does `BPF` have decimals? The number that is returned on the image is quite large. 

Here is a sample image the is returned:

download (1)


PatrickAlphaC:
This addresses https://github.com/BeanstalkFarms/Beanstalk/issues/120

PatrickAlphaC:
ALSO:

1. I'm not sure if I got the active, used, or available fertilizer math correct... hoping to get a double check and then update the tests
2. I did a weird swap-in-place bytecode trick to test the constant variable for the beanstalk diamond address. Let me know if that's not what you'd like to see.

PR #136, opened 2022-10-26
Update issue templates

Brean0:
A issue template standard is needed.  Ideally we should have a best practices for issues that mirrors the RFC. For example, we should adopt the modal language commonly used in technical documentation https://www.rfc-editor.org/rfc/rfc2119.

This is a suggested starting point, but very open to changes/additions, as it is very bare bones.


hellofromguy:
adding RFC/BIP templates in this PR: https://github.com/BeanstalkFarms/Beanstalk/pull/163

PR #141, opened 2022-11-10
BIP-X: Update Unripe Bean:3crv LP seeds from 4 to 2

pizzaman1337:
This is an early rough draft of a script to update the unripe lp seeds.

I'm new to Solidity, and don't have any blockchain projects in prod, so I expect there to be mistakes here and I'm mostly looking for some feedback, perhaps after a couple iterations I can get this up to a mergable state. I learned a lot from looking at other parts of the project, and several bits are copy/pasted here, such as from the replant code.

My concerns are:
1. The way I'm grabbing deposit info is by looking at all the `AddDeposit` events and filtering for Unripe LP. This works to get all the addresses and the seasons, which are the main things needed by the `LibTokenSilo.tokenDeposit()` function to get the _current_ deposit info on-chain. Only issue is, I suppose we could be wasting some gas by checking deposits that are already gone. Not sure what % of deposits that may be, but if it's high this could be a gas waste (perhaps it's a minor cost).
2. Is this the correct and best way of getting the deposit info and updating it on chain?
3. Each batch of 150 deposits takes a bit in my local `anvil` node, will it take that long on mainnet? If yes, then there's a chance the deposit state could change between when the script starts and when it finishes. I suppose one could always manually check for `AddDeposit` events after the block the script starts running on to clean up those stragglers.

I intentionally left some extra console logs in for now to help other people debug/understand what's happening. Sorry it's slightly messy in some places.

All feedback appreciated! I'd arb my unripe LP now while bean is under peg, but would rather hold for the seeds. My primary motivation for working on this is to help bean stay at peg, but also, I'm hoping to earn some beans from working on this BIP as well.


pizzaman1337:
After talking with publius and the BF team, updating seeds like this won't work since it will mess up the amount of stalk removed when a deposit is removed. To fix this, you'd need to move away from storing deposits in terms of seasons, and instead keep track of grown stalk on a per-deposit basis (which can be done through a shared index). I'm working on this and broader Silo changes here: https://github.com/BeanstalkFarms/Beanstalk/compare/zero-withdraw...pizzaman1337:Beanstalk:silov3#diff-a0abb2f4046ee1f0caff2716c2c0a21755aa0eac99ab5e6519fcb77d989a196cR199

PR #154, opened 2022-12-02
Tractor

beam00n:
https://github.com/BeanstalkFarms/Beanstalk/issues/290


protocol/contracts/libraries/LibFunction.sol
```diff
@@ -72,22 +72,30 @@ library LibFunction {
      @return data The function call return datas
     **/
     function buildAdvancedCalldata(
-        bytes calldata callData,
-        bytes calldata advancedData,
+        bytes memory callData,
+        bytes memory advancedData,
```
publiuss, line 76: It would be great to leave this as calldata to keep the gas costs of the standard farm version to a minimum. Do you have any thoughts on how to achieve this? Maybe add a `buildAdvancedCalldataMem` version?

funderberker:
@0xm00neth you should be able to save gas by using the signature as the TractorStorage nonce map key. This will avoid unnecessary duplicative hashing of the blueprint after the modifier has verified it already.

funderberker:
it would be possible to reduce the total number of times the blueprint hash must be calculated by including the hash in blueprint calldata and verifying it in the modifier along with the signature. unclear how much cost it would save and if it is worth the changes.

Issue #150, opened 2022-11-19
RFC: Instant withdraw 


Brean0:
## Summary
Remove the season lockup when withdrawing assets from the silo. 

## Problem
Currently there is a 1 Season lockup on Withdrawal. This provides an unnecessarily complex UX as Farmers are forced to perform 2 on-chain actions at different times. This limits the potential for using Deposited Beans throughout DeFi via Pipeline as Deposited Beans cannot atomically be swapped for Beans.

## Proposed Solution
Remove the Withdrawal Freeze entirely, allowing Farmers to Withdraw directly to Circulating/Farm balances in a single transaction.

### Technical implementation 
1. Modify `withdrawDeposit(s)` to send withdrawn tokens to Farm balance or Circulating balance using `sendToken` in `LibTransfer` instead of creating a Withdrawal. Will likely need to add reentrancy check to `withdrawDeposit(s)`.
2. Remove `claimDeposit(s)` as the function is no longer necessary.

### Additional info 
Currently, beans are issued at the start of the season. This creates an exploit where a farmer can deposit beans prior to the sunrise, call `sunrise()`, gain bean seigniorage, and immeditely sell beans with little to no bean exposure. 

Thus, this RFC will need to be implanted with/after the implementation of [Time Weighted Earned Beans Emissions ](https://github.com/BeanstalkFarms/Beanstalk/issues/149#issue-1456641055).

Issue #149, opened 2022-11-19
RFC: Delayed Earned Bean Seigniorage 


Brean0:
## Summary 
Delay the issuance of newly Earned Bean seigniorage after $n$ blocks, where $n$ is the number of blocks needed to mitigate multi-block MEV and flash loan attacks.  

## Problem
Beanstalk issues Earned Beans to Stalkholders at the `sunrise()` call. With the introduction of removing the withdraw timer, issuing beans at the start of the season introduces the following attacks: 

1.  With a flash loan, an attacker can deposit whitelisted assets into the silo, call `sunrise()`, withdraw instantly and gain risk-free seigniorage. 
2. With Multi-block MEV, an attacker can deposit whitelisted assets into the silo, call `sunrise()`, reject any blocks in which trades occur on AMMs paired with BEAN, and withdraw some blocks later, gaining risk-free seigniorage.

## Proposed Solution
Revise the bean issuance such that: 
1. Earned Beans of an given season is issued $n$ blocks after `sunrise()` is called.
2. Designate the period between the sunrise block and $n$ after to be called the vesting period.

In practice, this means: 
1. A farmer calling `plant()`in the vesting period would *defer* the beans issued until `plant()` is called outside the vesting period.
2. A farmer withdrawing their deposits (`withdrawDeposit(s)`) in the vesting period would *forfeit* their earned beans gained this season proportional to the withdrawn stalk amount to their total stalk amount (i.e if a user withdrew a deposit equal to 10% of their total stalk, they would forfeit 10% of the earned beans issued this season). 

### Technical Implementation 
1: split `earnedBeans` into two `uint128` variables: `earnedBeans` and `newEarnedStalk`. 
2: In `rewardToSilo`, set `s.newEarnedStalk = amount.mul(C.getStalkPerBean());`
3: Modify `_balanceOfEarnedBeans` to properly not include vesting Earned Stalk when computing Stalk including Earned Stalk.

### Technical Solution: 
In general, when a user mows their balances (assuming grown stalk > 0), stalk and roots are incremented to their account.  
The equation for calculating grown stalk and roots issued is: 

$$
\text{Grown Roots} =  \text{Roots}_S \cdot  \frac{\text{Grown Stalk}}{\text{Stalk}_S}  
$$

Where ${Roots}_S$ is the total Roots of the System and ${Stalk}_S$ is the total Stalk of the System. 

[...]

PR #172, opened 2022-12-12
BIP-X: Remove Withdraw Timer

Brean0:
# BIP-X: Remove Withdraw Timer

- [Proposer](#proposer)
- [Summary](#summary)
- [Problem](#problem)
- [Proposed Solution](#proposed-solution)

## Proposer

Beanstalk Farms

## Summary

- Adjust the issuance of earned Beans such that it is linearly distributed throughout the season, rather than at the start of the season.
- Remove the Withdrawal Freeze, allowing farmers to deposit and withdraw beans in the same season.
- Deprecate the various withdraw functions. 

## Problem

Numerous inefficiencies remain within Beanstalk: 

1. Beanstalk cannot currently support decrementing the Withdrawal timer to 0 because anyone could then receive “risk-free” Earned Beans because Earned Beans are distributed instantaneously when the sunrise() function is called. 
2. Having a withdraw lockup creates an unnecessarily complex UX as Farmers are forced to perform 2 on-chain actions at different times. This limits the potential for using Deposited Beans throughout DeFi via Pipeline as Deposited Beans cannot atomically be swapped for Beans.

## Proposed Solution

- Emit newly minted Earned Beans to Stalkholders proportional to the percent of the Season that has elapsed instead of instantly on the sunrise call. 
     - https://github.com/BeanstalkFarms/Beanstalk/issues/149
- Remove the Withdrawal Freeze entirely, allowing Farmers to Withdraw directly to Circulating/Farm balances in a single transaction.
     - https://github.com/BeanstalkFarms/Beanstalk/issues/150 

 
## Diagrams/Visualizations
image
image


protocol/contracts/beanstalk/farm/FarmFacet.sol
```diff
@@ -5,9 +5,9 @@
 pragma solidity ^0.7.6;
 pragma experimental ABIEncoderV2;
 
+import {LibDiamond} from "~/libraries/LibDiamond.sol";
+import {LibEth} from "~/libraries/Token/LibEth.sol";
```
funderberker, line 9: why did we change the import style and is it supposed to be changed to the tilde style everywhere?

protocol/contracts/beanstalk/init/replant/Replant1.sol
```diff
@@ -82,7 +82,7 @@ contract Replant1 {
 
     function init() external {
         // 1. Remove Deposits
-        LibTokenSilo.removeDeposit(
+        LibTokenSilo.removeDepositFromAccount(
```
funderberker, line 85: Do we need to keep these one time use legacy files up to date with changes? Seems best to keep them in the state that they were used in.

protocol/contracts/libraries/Silo/LibSilo.sol
```diff
@@ -8,98 +8,241 @@ pragma experimental ABIEncoderV2;
...
+        // We round up as it prevents an account having roots but no stalk.
+        
+        // if the user withdraws in the same block as sunrise, they forfeit their earned beans for that season
+        // this is distrubuted to the other users.
+        // should this be the same as the vesting period?
+        if(block.number - s.season.sunriseBlock <= 25){
+            roots = s.s.roots.mulDiv(
+            stalk,
+            s.s.stalk-s.newEarnedStalk,
+            LibPRBMath.Rounding.Up);
```
funderberker, line 224: @Brean0 previous logic was using safe math that rounds towards zero. why the change here and what are the implications?

protocol/contracts/beanstalk/AppStorage.sol
```diff
@@ -60,20 +69,50 @@ contract Account {
...
+         * 
+         * @dev FIXME(doc)
+         */
         uint32 votedUntil; // DEPRECATED – Replant removed on-chain governance including the ability to vote on BIPs.
         uint32 lastUpdate; // The Season in which the Farmer last updated their Silo.
         uint32 lastSop; // The last Season that a SOP occured at the time the Farmer last updated their Silo.
         uint32 lastRain; // The last Season that it started Raining at the time the Farmer last updated their Silo.
-        uint32 lastSIs; // DEPRECATED – In Silo V1.2, the Silo reward mechanism was updated to no longer need to store the number of the Supply Increases at the time the Farmer last updated their Silo.
-        uint32 proposedUntil; // DEPRECATED – Replant removed on-chain governance including the ability to propose BIPs.
+        uint128 deltaRoots; // the number of roots to add, in the case where a farmer has mowed in the morning
```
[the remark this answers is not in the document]
Brean0: in the EVM, data are stored in 32 byte slots. This allows for packing multiple variables into 1 slot, and adding more variables without overwriting the storage slot, as long the new variables do not exceed the 32 byte length. 

In this case, previously 6 uint32s were stored in this slot. This means that 24 bytes were used, and a potential 8 bytes can be added to this slot without affecting the leading slots after. 

by removing the last 2 uint32 variables in the slot, we can now pack a uint128 (16 bytes) into the slot without affecting the leading slots. The slot now looks like: 
uint32  (4 bytes)
uint32  (4 bytes)
uint32  (4 bytes)
uint32  (4 bytes) 
uint128 (16 bytes)
(4+4+4+4+12 = 32) 
additionally, because delta roots is only used in the vesting period, and is always updated before a vesting period plant, the previous values that was contained in the variables will be overwritten and not used.

protocol/contracts/beanstalk/AppStorage.sol
```diff
@@ -235,12 +294,12 @@ contract Storage {
...
 
 struct AppStorage {
-    uint8 index; // DEPRECATED - Was the index of the Bean token in the Bean:Eth Uniswap v2 pool, which has been depreciated.
+    uint8 index; // DEPRECATED - Was the index of the Bean token in the BEAN:ETH Uniswap v2 pool, which has been deprecated.
     int8[32] cases; // The 24 Weather cases (array has 32 items, but caseId = 3 (mod 4) are not cases).
     bool paused; // True if Beanstalk is Paused.
     uint128 pausedAt; // The timestamp at which Beanstalk was last paused. 
     Storage.Season season; // The Season storage struct found above.
-    Storage.Contracts c; // DEPRECATED - Previously stored the Contracts State struct. Removed when contract addresses were moved to constants in C.sol.
+    uint256[4] deprecated; // DEPRECATED - Previously stored the Contracts State struct. Removed when contract addresses were moved to constants in C.sol.
```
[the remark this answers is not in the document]
Brean0: since data is stored in 32 byte chunks, `address[4]`  still used 4 slots of storage, even though the array technically only used (4 * 20) 80 bytes. thus 4 slots are deprecated here (uint256[4] and bytes32[4] are the same in this case)

protocol/contracts/beanstalk/AppStorage.sol
```diff
@@ -284,6 +358,7 @@ struct AppStorage {
     uint128 fFirst; // The lowest active Fertilizer Id (start of linked list that is stored by nextFid). 
     uint128 fLast; // The highest active Fertilizer Id (end of linked list that is stored by nextFid). 
     uint128 bpf; // The cumulative Beans Per Fertilizer (bfp) minted over all Season.
+    uint128 newEarnedRoots; // the number of roots to add to the global roots, in the case the user plants in the morning. // placed here to save a storage slot.s
```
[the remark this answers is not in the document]
Brean0: bpf slot previously: 
uint128 (16 bytes) 
16 free bytes -> newEarnedRoots

protocol/contracts/libraries/Silo/LibSilo.sol
```diff
@@ -8,98 +8,241 @@ pragma experimental ABIEncoderV2;
...
+
         uint256 roots;
-        if (s.s.roots == 0) roots = stalk.mul(C.getRootsBase());
-        else roots = s.s.roots.mul(stalk).div(s.s.stalk);
+        if (s.s.roots == 0) {
+            roots = uint256(stalk.mul(C.getRootsBase()));
+        } else  {
+            roots = s.s.roots.mul(stalk).div(s.s.stalk);
+            if (block.number - s.season.sunriseBlock <= 25) {
+                uint256 rootsWithoutEarned = s.s.roots.add(s.newEarnedRoots).mul(stalk).div(s.s.stalk - (s.newEarnedStalk));
```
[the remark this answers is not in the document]
Brean0: `newEarnedRoots` are not accounted for in s.s.roots, as roots are not incremented when new beans are minted, but rather during the mow function.

protocol/contracts/libraries/Silo/LibSilo.sol
```diff
@@ -8,98 +8,241 @@ pragma experimental ABIEncoderV2;
...
-        uint256 roots = s.s.roots.mul(stalk).div(s.s.stalk);
+       
+        uint256 roots;
+        // Calculate the amount of Roots for the given amount of Stalk.
+        // We round up as it prevents an account having roots but no stalk.
+        
+        // if the user withdraws in the same block as sunrise, they forfeit their earned beans for that season
+        // this is distrubuted to the other users.
+        // should this be the same as the vesting period?
+        if(block.number - s.season.sunriseBlock <= 25){
```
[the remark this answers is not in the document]
Brean0: agree

protocol/contracts/libraries/Silo/LibSilo.sol
Brean0: previously the roots here was truncated - with the introduction of more complex math, this adds an additional fail safe to prevent rounding errors. This follows defi standards (we round down when issuing roots, round up when removing roots) to prevent the protocol from being undercollateralized

Brean0: Additionally, line 95 adds an additional precaution to prevent underflow.

protocol/contracts/libraries/Silo/LibTokenSilo.sol
```diff
@@ -25,172 +34,266 @@ library LibTokenSilo {
...
+     *
+     * FIXME(discuss): If we think of Deposits like 1155s, we might call the
+     * combination of "incrementTotalDeposited" and "addDepositToAccount" as 
+     * "minting a deposit".
+     */
     function depositWithBDV(
         address account,
         address token,
-        uint32 _s,
+        uint32 season,
```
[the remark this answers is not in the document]
Brean0: season >>>> _s

protocol/contracts/beanstalk/AppStorage.sol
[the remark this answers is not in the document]
funderberker: got it. ty for the explanation

protocol/contracts/beanstalk/silo/SiloFacet/SiloExit.sol
```diff
@@ -0,0 +1,333 @@
...
+        view
+        returns (uint256 beans) {
+        // There will be no Roots before the first Deposit is made.
+        if (s.s.roots == 0) return 0;
+
+        uint256 stalk;
+        if(block.number - s.season.sunriseBlock <= EARNED_BEAN_VESTING_BLOCKS){
+            stalk = s.s.stalk.sub(s.newEarnedStalk).mulDiv(
+                s.a[account].roots.add(s.a[account].deltaRoots), // add the delta roots of the user
+                s.s.roots.add(s.newEarnedRoots), // add delta of global roots
```
funderberker, line 192: s.newEarnedBeans is set when calling mow, which may not be called before checking of balance here. should we trigger an update before this addition?

protocol/contracts/libraries/Silo/LibSilo.sol
funderberker: what is the precaution? line 95 is `mintStalk(account, stalk); // also mints Roots`

funderberker: let's talk about rounding on the next call. i think i am missing a lot of context

protocol/contracts/libraries/Silo/LibSilo.sol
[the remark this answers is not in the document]
funderberker: ok, but why are they being added here? this logic is (roots with earned) * (stalk ratio without earned)
but seems like it should be (roots without earned) * (stalk ratio without earned)

protocol/contracts/libraries/Silo/LibSilo.sol
```diff
@@ -8,98 +8,241 @@ pragma experimental ABIEncoderV2;
...
-        if (s.s.roots == 0) roots = stalk.mul(C.getRootsBase());
-        else roots = s.s.roots.mul(stalk).div(s.s.stalk);
+        if (s.s.roots == 0) {
+            roots = uint256(stalk.mul(C.getRootsBase()));
+        } else  {
+            roots = s.s.roots.mul(stalk).div(s.s.stalk);
+            if (block.number - s.season.sunriseBlock <= 25) {
+                uint256 rootsWithoutEarned = s.s.roots.add(s.newEarnedRoots).mul(stalk).div(s.s.stalk - (s.newEarnedStalk));
+                uint256 deltaRoots = rootsWithoutEarned - roots;
+                s.newEarnedRoots = s.newEarnedRoots.add(uint128(deltaRoots));
```
funderberker, line 160: this `deltaRoots` is the delta for the account stalk amount entered as parameters here. is it intentional that the silo state `newEarnedRoots` is holding a value for an account?

protocol/contracts/libraries/Silo/LibSilo.sol
```diff
@@ -8,98 +8,241 @@ pragma experimental ABIEncoderV2;
...
+                s.a[account].deltaRoots = uint128(deltaRoots);
+            } 
+        }
 
+        // increment user and total stalk
         s.s.stalk = s.s.stalk.add(stalk);
         s.a[account].s.stalk = s.a[account].s.stalk.add(stalk);
 
+        // increment user and total roots
         s.s.roots = s.s.roots.add(roots);
```
funderberker, line 170: this roots incrementation includes "new earned roots" is that intentional?

protocol/test/Silo.test.js
```diff
@@ -133,5 +144,302 @@ describe('Silo', function () {
       await this.silo.connect(user2).plant()
       expect(await this.silo.totalEarnedBeans()).to.eq('0');
     });
-  })
+  });
+  
+
+  describe("Time Weighted Earned Bean Emission", async function () {
```
funderberker, line 150: reminder: update name

protocol/contracts/beanstalk/init/replant/Replant1.sol
funderberker: @silochad, do you have an opinion?

silochad:
Closing as this has been manually merged into `grown-stalk-per-bdv`.

PR #181, opened 2023-01-06
Add dev graph node project

marshall-bean:
Add a new project to the monorepo to setup a local graph-node instance.


0xalecks (approved):

silochad:
Should we expand the scope of this one to include deploy scripts for `subgraph-beanstalk` and `subgraph-bean` to the node after it's instantiated?

PR #239, opened 2023-01-06
Monorepo: Update typedef generation

0xalecks:
This PR makes the following changes:

- makes /protocol an package that can be imported, it only exports "/abi" path
- SDK will generate typedefs by referencing the Beanstalk ABI from /protocol
- UI will generate typedefs by referencing the Beanstalk ABI from /protocol
- UI will no longer directly store Beanstalk ABI, but instead import it
 ex: `import BEANSTALK_ABI from '@beanstalk/protocol/abi/Beanstalk.json';`
- All projects have a local `yarn generate` script, that builds or generates whatever assets
- Protocol has a `yarn generate` script that compiles, then builds the concatenated ABI (hardhat diamondABI).
- root level scripts added for each project, ex: `yarn ui:generate`
- root level `yarn bootstrap` added that install husky and runs global generate
- fixes issues created by the new typechain generation (the generated assets now have a different folder struction due to the expanded glob passed to typechain command)
- re-enables husky and adds a post-update script that just shows a tip


silochad:
Reorganized the top-level package.json commands with this structure:

```
Order:
- monorepo
- protocol
- sdk
- ui
- other

Within:
- setup (bootstrap, generate)
- dev
- build
- test
- release (format, version, publish)
```

silochad (approved):

0xalecks:
reviewed latest commit, LGTM

Issue #182, opened 2023-01-05
Pod Market rounding and filtering


silochad:
- Beans Remaining is > 0 but rounds down to 0
- Pods Requested rounds up to 0.04 from 0.0379
- This Order is displayed even though it has ~0 Beans left.

![image](https://user-images.githubusercontent.com/99763686/210830149-fb43c4fe-0ffa-43d8-9155-38339d77f62e.png)
![image](https://user-images.githubusercontent.com/99763686/210830208-8d2d0db1-005a-45da-9ab9-54185c344489.png)

PR #242, opened 2023-01-06
fix: pod market rounding & filtering

imcoolbean:


projects/ui/src/hooks/beanstalk/useMarketData.ts
```diff
@@ -20,7 +22,8 @@ const useMarketData = () => {
 
   /// Cast query data to BigNumber, etc.
   const listings = useCastApolloQuery<PodListing>(listingsQuery, 'podListings', useCallback((_listing) => castPodListing(_listing, harvestableIndex), [harvestableIndex]), loading);
-  const orders   = useCastApolloQuery<PodOrder>(ordersQuery, 'podOrders', castPodOrder, loading);
+  let orders   = useCastApolloQuery<PodOrder>(ordersQuery, 'podOrders', castPodOrder, loading);
+  orders = orders?.filter((order) => order.beanAmountRemaining.gt(MIN_POD_AMOUNT));
```
imcoolbean, line 26: @silochad since `beanAmountRemaining` is something we calculate, not sure if the filter is possible during the query

silochad: This is a great point. @cujowolf In order to let users query by the number of Pods remaining in an order, they'll need an explicit field on the entity (AFAIK there is no way to query where `beanAmount - beanAmountFilled > X` for example.

silochad:
Checks failing because build error fix is currently on master, should deploy correctly upon merge.

imcoolbean:
Fixes #182

PR #248, opened 2023-01-10
monorepo: tests workspace

Bean-Sama:
- set up tests workspace
- create basic ui test that checks for navigation using cypress

Note: It may be a better idea to use Vitest for env vars (need to impersonate wallet for txns), but not sure yet what the best approach is


projects/tests/babel.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
```
0xalecks, line 1: Should we source from the root babel config first, then overwrite as needed?:
```
{
   "extends": "../../babel.config.json",
   ...whatever here..
}
```
( #239 converts the babel config to json )

0xalecks (approved):

silochad (approved):

projects/tests/babel.config.js
Bean-Sama: added this!

projects/ui/src/components/Nav/Buttons/LinkButton.tsx
```diff
@@ -60,6 +60,7 @@ const LinkButton: FC<{ to: string; title: string, tag?: string }> = ({ to, title
             display: 'inline-block',
             textDecorationThickness: '2px',
           }}
+          data-cy={`Navbar-${title}`}
```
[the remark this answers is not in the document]
Bean-Sama: added a vite plugin to remove these on prod

PR #277, opened 2023-01-15
Claim and Do X components

Bean-Sama:


silochad:
Changes requested from my first review:

-[ ] Check Discord thread for action items and drop a ✅ when completed
-[ ] Constrain the right column to a fixed width, right now it expands if you type large numbers that result in large outputs. In this screenshot, the side-by-side TokenOutputs are causing the issue.
image

projects/ui/src/components/Common/SelectionIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+      {selected ? (
+        <Check
+          sx={{
+            color: BeanstalkPalette.white,
+            width: size * SIZE_MULTIPLIER,
+            height: 'auto',
+          }}
+        />
+      ) : null}
+    </Stack>
```
silochad, line 49: Looks like it's just the check icon in the top right? This strikes me as a lot of code for just the icon. Could it be reduced down to a CheckCircle with a set color?

https://mui.com/material-ui/material-icons/?query=check&selected=CheckCircle

projects/ui/src/components/Common/Form/TokenSelectDialog.tsx
```diff
@@ -85,6 +114,13 @@ const TokenSelectDialog : TokenSelectDialogC = React.memo(({
     }
   }, [selectedInternal]);
 
+  const filteredTokenList = useMemo(() => {
```
silochad, line 117: Add comment: `// ETH can only be used from EXTERNAL balance`

projects/ui/src/components/Common/Form/TokenInputField.tsx
```diff
@@ -52,6 +57,10 @@ export type TokenInputCustomProps = {
    *
    */
   allowNegative?: boolean;
+  /**
+   * whether to use the wrapped variant or the default text field variant
+   */
+  inputVariant?: 'default' | 'wrapped'
```
silochad, line 63: Do we need to add support for the old variant? Wondering if we can just pull this and have the new `wrapped` version propagate across the site.

silochad:
- [ ] Bug: in this screenshot I have 491 BEAN in my Circulating balance but it appears when I select the Farm mode
image

silochad (changes requested):

projects/ui/src/components/Common/SelectionIndicator.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { remToPx } from '~/util/UI';
+
+export type SelectionIndicatorProps = {
+  selected: boolean;
+  size?: keyof typeof FontSize | number;
+} & BoxProps;
+
+const SIZE_MULTIPLIER = 0.75;
+
+const SelectionIndicator: React.FC<SelectionIndicatorProps> = ({
```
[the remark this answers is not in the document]
Bean-Sama: just removed this component all together

projects/ui/src/components/Common/Form/TokenInputField.tsx
Bean-Sama: Yup! Can do!

silochad:
In some cases, looks like the select dropdown button is off-center:
image

Issue #267, opened 2023-01-13
Add horizontal line at 0 on deltaB chart


silochad:
![image](https://user-images.githubusercontent.com/99763686/212382324-81671cfb-1e6b-4e82-9c7f-3505bf2e77ab.png)

k-xo:
Would love to take a stab at this will open a PR soon!

PR #280, opened 2023-01-16
Add Zero Horizontal Line on Delta B Chart

k-xo:
Addresses #267


projects/ui/src/components/Common/Charts/ChartPropProvider.tsx
```diff
@@ -135,6 +135,7 @@ export type BaseChartProps = {
   curve?: CurveFactory | keyof typeof CURVES;
   scale?: keyof typeof SCALES;
   isTWAP?: boolean;
+  isZeroHorizontal?: boolean;
```
silochad, line 138: @k-xo Can you walk me thru why you chose this route instead condensing both into one var like `horizontalLine: number`?

projects/ui/src/components/Common/Charts/MultiLineChart.tsx
```diff
@@ -266,7 +267,24 @@ const MultiLineChart: React.FC<BaseChartProps> = (props) => (
             {...providerProps}
             {...props}
           >
-            {(childProps) => <ExploitLine {...childProps} />}
+            {(childProps) => (
+              <>
+                <ExploitLine {...childProps} />
+                {props.isZeroHorizontal && (
+                  <Line
+                    from={{ x: 0, y: childProps.scales[0].yScale(1) as number }}
```
silochad, line 275: Believe this will put a value at `y = 1` instead of `y = 0`

silochad (changes requested):

projects/ui/src/components/Common/Charts/ChartPropProvider.tsx
k-xo: Ah yes, was getting the zeroHorizontal approach to work, but as mentioned final version would be better with a generalised version, will make this change now

silochad (approved):

silochad:
Thank you @k-xo! 🚀

Issue #185, opened 2023-01-02
Analytics page deep linking


silochad:
Re-add deep links on the analytics page (ex. https://app.bean.money/#/analytics?field=sowers)

silochad:
Resolved in #281

PR #281, opened 2023-01-16
Re-adds deep linking to analytics page

k-xo:
Addresses #185


projects/ui/src/components/Analytics/Bean/index.tsx
```diff
@@ -9,12 +9,12 @@ import Volume from '~/components/Analytics/Bean/Volume';
 import Liquidity from '~/components/Analytics/Bean/Liquidity';
 import Crosses from '~/components/Analytics/Bean/Crosses';
 import DeltaB from '~/components/Analytics/Bean/DeltaB';
-
-// const SLUGS = ['price', 'volume', 'liquidity', 'mktcap', 'supply', 'crosses'];
 import { FC } from '~/types';
 
+const SLUGS = ['price', 'volume', 'liquidity', 'mktcap', 'supply', 'crosses', 'DeltaB'];
```
silochad, line 14: Instead of removing the space (which is important cosmetically in the UI), could you suggest a different approach that lets a dev add `label="Delta B"` but but keeps the slug url-safe?

I can think of two satisfactory approaches, let me know what you think of.

projects/ui/src/components/Analytics/Field/index.tsx
```diff
@@ -9,22 +9,34 @@ import Temperature from './Temperature';
...
+        sx={{ px: 2, pt: 2, pb: 1.5 }}
+      >
         <Tab label="RRoR" />
         <Tab label="Temperature" />
         <Tab label="Pods" />
         <Tab label="Pod Rate" />
         <Tab label="Sown" />
         <Tab label="Harvested" />
-        <Tab label="Total Sowers" />
+        <Tab label="TotalSowers" />
```
silochad, line 39: See above comments, we'll want to keep this as "Total Sowers" for cosmetic & readability reasons. (The `label` is what's actually disabled for the user to click on)

projects/ui/src/components/Analytics/Silo/index.tsx
```diff
@@ -9,15 +14,28 @@ import Seeds from '~/components/Analytics/Silo/Seeds';
...
+  'deposited_urlp',
+  'stalk',
+  'seeds',
+  'bean_vAPY',
+  'lp_vAPY',
+];
+
 const SiloAnalytics: FC<{}> = () => {
-  const [tab, handleChangeTab] = useTabs();
+  const [tab, handleChangeTab] = useTabs(SLUGS, 'deposits');
```
silochad, line 31: This tab refers to analytics about the Silo, so would suggest using `silo` instead of `deposits`

projects/ui/src/components/Analytics/Bean/index.tsx
```diff
@@ -9,12 +9,12 @@ import Volume from '~/components/Analytics/Bean/Volume';
...
+
 const BeanAnalytics: FC<{}> = () => {
-  const [tab, handleChangeTab] = useTabs();
+  const [tab, handleChangeTab] = useTabs(SLUGS, 'lens');
```
silochad, line 17: This tab refers to analytics about the Bean token, would suggest using `bean` instead of `lens`

silochad (changes requested):
I remembered the issue with the `useTabs()` implementation which caused me to remove it. It has to do with a single page utilizing multiple `useTabs()` hooks. It's present in this PR too -- see if you can find it.

silochad (approved):

silochad:
Changes look great

PR #287, opened 2023-01-19
Use SDK For Swap

marshall-bean:
Replace the logic in the swap page with the SDK.

- Removed unnecessary logic that the SDK has now
- Fixed UI CI
- Fix UI build command (e.g. used by netlify deploy)
- Bump vite to 3.2.5 to replace the HMR fix from @0xalecks 's original PR https://github.com/BeanstalkFarms/Beanstalk-UI/pull/518. It seemed to fix it.

Manually tested many scenarios such as:
- Swap BEAN/ETH with "forward" quote
- Swap ETH/3CRV (should show "not supported" message
- Reverse quote (type in the "destination" the amount out)
- Click reverse button and get expected results (the button to swap source/destination input)
- Move BEAN from wallet (circulating) to farm balance and back
- Swap BEAN/ETH with amount of BEAN that exceeds farming balance


projects/ui/src/graph/graphql.schema.json
```diff
@@ -44754,7 +44754,7 @@
           },
           {
             "name": "index",
-            "description": "The Pod Index of the listed Plot.",
+            "description": "The absolute index of the listed Plot in the Pod Line.\n\nMeasured from the front, so the Listing contains all Pods between\n(index) and (index + totalAmount).\n\nAn example where the podLine is 50,000 but the index is 150,000:\n   0         the first Pod issued\n   100,000   harvestableIndex\n   150,000   index\n",
```
marshall-bean, line 44757: I'm hoping someone can provide some context about why this change happened, presumably after `yarn generate`.

0xalecks: yup. if you follow the yarn generate command, you'll see it does a graphql-codegen, which pulls this schema from a hardcoded url. Must've changed upstream

marshall-bean: OK just making sure it's expected and it's fine to commit. Thanks.

projects/ui/src/components/Swap/Actions/Swap.tsx
```diff
@@ -667,197 +713,31 @@ const Swap: FC<{}> = () => {
...
-        _tokenIn,
-        _tokenOut,
-        FarmFromMode.INTERNAL_EXTERNAL,
-        FarmToMode.EXTERNAL,
-      );
+
+      const forward: Boolean = direction === 'forward';
+
+      const amountIn = forward ? 
+        ethers.BigNumber.from(toStringBaseUnitBN(_amountIn, swapOperation.tokenIn.decimals))
```
0xalecks, line 728: Think these conversions can be simplified. The UI uses a mix of ethers.BigNumber and BigNumber.js along with a bunch of custom conversion and formatting methods, like toStringBaseUnitBN.

All of this was simplified in the SDK. TokenValue is the all encompassing class to use for values.

Would be good to replace all usage of BigNumber with TokenValue as part of this effort.

For ex this code could be:
```
     const amountIn = forward ?
        swapOperation.tokenIn.fromBlockchain( _amountIn.toString())
        : swapOperation.tokenOut.fromBlockchain( _amountIn.toString());
```  
This makes `amountIn` a TokenValue, which is accepted by the `swapOperation.estimate()` method. 

Even further, the `_amountIn` variable here is BigNumber.js. Tracing where that is created and making it a TokenValue there would be ideal.

projects/ui/src/components/Swap/Actions/Swap.tsx
```diff
@@ -143,18 +149,46 @@ const SwapForm: FC<FormikProps<SwapFormValues> & {
...
+    if (!sdkTokenIn) {
+      throw new Error(`Address of ${uiTokenIn.symbol} was not found in SDK tokens.`);
+    }
+    const sdkTokenOut = sdk.tokens.findByAddress(uiTokenOut.address);
+    if (!sdkTokenOut) {
+      throw new Error(`Address of ${uiTokenOut.symbol} was not found in SDK tokens.`);
+    }
+
+    return sdk.swap.buildSwap(sdkTokenIn, sdkTokenOut, account!, farmFrom, farmTo);
+  }, [sdk.tokens, sdk.swap, account]);
```
0xalecks, line 169: Having sdk.tokens and sdk.swap here is very strange. I know this is to make [react-hooks/exhaustive-deps] happy, but feels so wrong, lol

projects/ui/src/components/Swap/Actions/Swap.tsx
0xalecks: FYI, just chatted with @silochad about this, switching to TokenValue all the way through would involve forking the TokenInput component, so as to not break all other forms that share it. And there's parallel work being done on these so we'll defer till later.

marshall-bean: Got it, will look into this.

marshall-bean: OK 👍

silochad (approved):

PR #291, opened 2023-01-24
Claim and do x integration

Bean-Sama:


projects/ui/src/components/App/SdkProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+    _sdk.tokens.UNRIPE_BEAN.setMetadata({ logo: UNRIPE_BEAN[1].logo });
+    _sdk.tokens.UNRIPE_BEAN_CRV3.setMetadata({ logo: UNRIPE_BEAN_CRV3[1].logo });
+
+    _sdk.tokens.SPROUTS.setMetadata({ logo: SPROUTS.logo });
+    _sdk.tokens.PODS.setMetadata({ logo: PODS.logo });
+    
+    _sdk.tokens.DAI.setMetadata({ logo: DAI[1].logo });
+    _sdk.tokens.USDC.setMetadata({ logo: USDC[1].logo });
+    _sdk.tokens.USDT.setMetadata({ logo: USDT[1].logo });
+    _sdk.tokens.CRV3.setMetadata({ logo: CRV3[1].logo });
```
silochad, line 52: @0xalecks: if we go this route, don't use the old token objects as a reference for `.logo`, imagine that old tokens are going to get removed.

projects/ui/src/components/App/SdkProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+      provider = undefined;
+    }
+
+    const _provider = new ethers.providers.WebSocketProvider(
+      'ws://localhost:8545'
+    );
+
+    const _sdk = new BeanstalkSDK({
+      signer: signer ?? undefined,
+      provider: _provider,
```
marshall-bean, line 35: FYI here's my implementation https://github.com/BeanstalkFarms/Beanstalk/pull/287/files#diff-a694b8b0c305431e644cd266565c1ac367fde96fd5066815cc39284c2c7d21f1 

Granted, I might be wrong but I believe provider is automatically set to `signer.provider`.

PR #299, opened 2023-01-25
SDK: Silo withdraw

0xalecks:
adds support for silo withdraws


projects/sdk/src/lib/silo/Withdraw.ts
```diff
@@ -0,0 +1,128 @@
...
+   * from, and therefore is limited to only whitelisted assets.
+   * Behind the scenes, the `amount` to be withdrawn must be taken from individual
+   * deposits, aka crates. A user's deposits are not summarized into one large bucket, from
+   * which we can withdraw at will. Each deposit is independently tracked, so each withdraw must
+   * calculate how many crates it must span to attain the desired `amount`.
+   * @param token The whitelisted token to withdraw. ex, BEAN vs BEAN_3CRV_LP
+   * @param amount The desired amount to withdraw. Must be 0 < amount <= total deposits for token
+   * @returns Promise of Transaction
+   */
+  async withdraw(token: Token, amount: TokenValue): Promise<ContractTransaction> {
```
marshall-bean, line 26: I think another method name might work better. Right now, you end up calling `sdk.silo.withdraw.withdraw` and it's weird, but other entities build the entity then `.execute()`, e.g. `sdk.silo.buildDeposit` then `deposit.execute()`. Similarly with swaps. Does that pattern not make sense here for some reason?

projects/sdk/src/constants/addresses.ts
```diff
@@ -10,7 +10,7 @@ export const addresses = {
   // ----------------------------------------
   // Ecosystem Contracts
   // ----------------------------------------
-  BEANSTALK_PRICE: Address.make("0xA57289161FF18D67A68841922264B317170b0b81"),
+  BEANSTALK_PRICE: Address.make("0xF2C2b7eabcB353bF6f2128a7f8e1e32Eeb112530"),
```
marshall-bean, line 13: Why this change?

projects/sdk/src/lib/silo/Withdraw.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    return contractCall;
+  }
+
+  /**
+   * This methods figures out which deposits, or crates, the withdraw must take from
+   * in order to reach the desired amount. It returns extra information that may be useful
+   * in a UI to show the user how much stalk and seed they will forfeit as a result of the withdraw
+   */
+  calculateWithdraw(token: Token, amount: TokenValue, crates: DepositCrate[], season: number) {
```
marshall-bean, line 69: What's the use case for this (and the pickCrates) being publicly accessible? Do we surface that in the UI somewhere? If so, might make sense to add to `examples`.

marshall-bean (changes requested):

projects/sdk/src/lib/silo/utils.ts
```diff
@@ -0,0 +1,47 @@
+import { Token } from "src/classes/Token";
+import { TokenValue } from "src/TokenValue";
+import { DepositCrate, Silo } from "../silo";
+
+/**
+ * Selects the number of crates needed to add up to the desired `amount`.
+ */
+export function pickCrates(crates: DepositCrate[], amount: TokenValue, token: Token, currentSeason: number) {
```
marshall2112, line 8: I'd add unit tests for this function.

projects/sdk/src/lib/silo/Withdraw.test.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Token } from "src/classes/Token";
+import { TokenValue } from "src/TokenValue";
+import { getTestUtils } from "src/utils/TestUtils/provider";
+import { Withdraw } from "./Withdraw";
+
+const { sdk, account, utils } = getTestUtils();
+
+jest.setTimeout(30000);
+
+describe("Silo Withdrawl", function () {
```
marshall2112, line 12: After working on Transfer, I think there's value in creating two test cases - sourcing a withdraw from single crate vs multiple crates, to make sure the SDK handles both appropriately.

projects/sdk/src/lib/silo/Withdraw.ts
```diff
@@ -0,0 +1,69 @@
+import { ContractTransaction } from "ethers";
+import { Token } from "src/classes/Token";
+import { TokenValue } from "src/classes/TokenValue";
+import { BeanstalkSDK } from "../BeanstalkSDK";
+import { DepositCrate, Silo } from "../silo";
+import { sortCratesBySeason } from "../silo.utils";
```
marshall2112, line 6: Why do we have both `silo.utils` and `./utils`? What's the difference between these two?

marshall2112 (changes requested):

projects/sdk/src/lib/silo/Withdraw.ts
marshall-bean, line 6: Why do we have both silo.utils and ./utils? What's the difference between these two?

projects/sdk/src/lib/silo/utils.ts
marshall-bean, line 8: I'd add unit tests for this function.

projects/sdk/src/lib/silo/Withdraw.test.ts
marshall-bean, line 12: After working on Transfer, I think there's value in creating two test cases - sourcing a withdraw from single crate vs multiple crates, to make sure the SDK handles both appropriately.

marshall-bean (changes requested):

projects/sdk/src/lib/silo/Withdraw.ts
0xalecks: it's temporary until we do silo cleanup. Utils should go into the /silo subfolder, so that's why I added that there. On cleanup, the silo.utils stuff should move into /silo/utils.ts

projects/sdk/src/lib/silo/Withdraw.test.ts
0xalecks: I tried, but this requires forwarding the season twice to get the setup, and this caused the whole test case to run beyond the 30 seconds default.

projects/sdk/src/lib/silo/utils.ts
0xalecks: good callout

projects/sdk/src/lib/silo/Withdraw.test.ts
marshall-bean: You only need to forward _once_ don't you? Deposit in season n, ffwd to season n+1, deposit again. Your withdrawals could then be sourced from two crates.

marshall-bean (approved):

PR #300, opened 2023-01-26
Remove pathway dependency from swap

silochad:
The "pathway" system on the Swap page was previously used to identify a swap route between two tokens. A limited number of swap routes were supported: for example the ETH <-> BEAN pair had a route, but USDC <-> DAI did not.

The new Swap router in the SDK handles this internally, and is able to arbitrarily route between tokens as long as an appropriate step is created which defines how to move along a particular token -> token edge. 

Removing the pathway system allows for new swap paths to work out of the box. Screenshot below shows the previously unsupported stable <-> stable pathway.

This PR:
- [x] removes the pathway system 
- [x] fixes bug with slippage (division by 100 was moved into SDK)

image

### DIFF
--- a/projects/ui/src/components/Swap/Actions/Swap.tsx
+++ b/projects/ui/src/components/Swap/Actions/Swap.tsx
@@ -715,7 +715,7 @@ const Swap: FC<{}> = () => {
           success: 'Swap successful.'
         });
 
-        const txn = await values.swapOperation.execute(amountIn, values.settings.slippage / 100);
+        const txn = await values.swapOperation.execute(amountIn, values.settings.slippage);
         txToast.confirming(txn);
 
         const receipt = await txn.wait();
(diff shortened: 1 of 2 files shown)


projects/ui/src/components/Swap/Actions/Swap.tsx
```diff
@@ -465,7 +447,7 @@ const SwapForm: FC<FormikProps<SwapFormValues> & {
               /// user has no balance of the input.
               || noBalance
               /// No way to quote for this pathway
-              || !pathwayCheck
+              // || !pathwayCheck
```
marshall2112, line 450: Nit: We can just remove it instead of commenting it out

projects/ui/src/components/Swap/Actions/Swap.tsx
```diff
@@ -417,7 +399,7 @@ const SwapForm: FC<FormikProps<SwapFormValues> & {
             }
             disabled={
               quotingIn
-              || !pathwayCheck
+              // || !pathwayCheck
```
marshall2112, line 402: Nit: We can just remove it instead of commenting it out

marshall2112 (approved):
LGTM aside from two nits, neither of which are a big deal.

projects/ui/src/components/Swap/Actions/Swap.tsx
silochad: Great, will leave for now and remove on next edit.

PR #301, opened 2023-01-27
SDK: Silo Claim

0xalecks:
## PR currently targets branch `silo-withdraw`. If that branch is merged first (#299 ), this PR needs to be updated to retarget master

Adds:
- sdk.silo.claim()
- sdk.silo.claimSeasons()
- sdk.silo.getClaimableAmount()

### DIFF
--- a/projects/sdk/src/lib/silo.ts
+++ b/projects/sdk/src/lib/silo.ts
@@ -17,6 +17,7 @@ import { assert } from "src/utils";
 import { DepositBuilder } from "./silo/DepositBuilder";
 import { DepositOperation } from "./silo/DepositOperation";
 import { Withdraw } from "./silo/Withdraw";
+import { Claim } from "./silo/Claim";
 
 /**
  * A Crate is an `amount` of a token Deposited or
@@ -82,6 +83,7 @@ export class Silo {
   static sdk: BeanstalkSDK;
   private depositBuilder: DepositBuilder;
   withdraw: Withdraw;
+  claim: Claim;
   // 1 Seed grows 1 / 10_000 Stalk per Season.
   // 1/10_000 = 1E-4
   // FIXME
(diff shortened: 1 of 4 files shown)


projects/sdk/src/lib/silo/Claim.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /**
+   * Claims specific seasons from Silo claimable amount.
+   * @param token Which Silo token to withdraw. Must be a whitelisted token
+   * @param seasons Which seasons to claim, from the available claimable list. List of seasons
+   * can be retrieved with .getClaimableAmount()
+   * @param toMode Where to send the output tokens (circulating or farm balance)
+   */
+  async claimSeasons(token: Token, seasons: string[], toMode: FarmToMode = FarmToMode.EXTERNAL) {
```
silochad, line 50: I think we should have a higher-level discussion about naming variance between the SDK <> contract, but think this format (`claim` = claim all, `claimSeasons` = granular control) makes sense. Can discuss after Claim functionality is removed in zero withdraw BIP.

projects/sdk/src/lib/silo.ts
```diff
@@ -91,6 +93,7 @@ export class Silo {
     Silo.sdk = sdk;
     this.depositBuilder = new DepositBuilder(sdk);
     this.withdraw = new Withdraw(sdk);
+    this.claim = new Claim(sdk);
```
silochad, line 96: Expanding on the `sdk.silo.claim.claim` note, feels strange to me to have `depositBuilder` (with "builder") but `withdraw` and `claim`. Understand though that the DepositBuilder is doing something functionally different, just thinking about this from an interface perspective.

silochad (approved):

projects/sdk/src/lib/silo/Claim.ts
0xalecks: sounds good

projects/sdk/src/lib/silo.ts
0xalecks: Yeah, it is weird, and it's something that bugged me too. The reason for depositBuilder, and swapBuilder, is to optimize in the UI. It gives you an object that has route calculated and allows you to play with the amounts w/o rebuilding the whole thing. We can discuss if there's way to improve this or if you think it's over-optimized.

PR #329, opened 2023-02-07
Wells: SDK foundation, layout

0xalecks:
- extracts TokenValue, Token, and some other classes into a stand alone new package: @beanstalk/sdk-core
- adds new sdk-wells package, with barebones structure. builds successfully
- examples for core and wells sdks, confirming compile/build/etc steps work


projects/examples/package.json
```diff
@@ -3,14 +3,16 @@
   "main": "src/index.ts",
   "dependencies": {
     "@beanstalk/sdk": "workspace:*",
+    "@beanstalk/sdk-core": "workspace:*",
```
marshall-bean, line 6: Minor thing, but to me "core" implies "foundation" when really these are just common classes, so I'd suggest "sdk-common" as the package name instead.

projects/sdk-core/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.log
+.DS_Store
+node_modules
+dist
+cache
+.env
```
marshall-bean, line 6: We need to ignore `src/constants/generated/**/*` and `src/constants/generated-gql/**/*` also don't we? The `generate` command in sdk-core would leave artifacts there.

projects/sdk-core/src/constants/abi/ERC20.json
```diff
@@ -0,0 +1,117 @@
+[
```
marshall-bean, line 1: Should these and other src/constants be committed?

projects/examples/package.json
```diff
@@ -3,14 +3,16 @@
...
+    "@beanstalk/wells": "workspace:*",
```
marshall-bean, line 7: Adding this here but applies to `jest.config.ts` - should we add these new projects there? Along with top-level sdk-core and sdk-wells package.json commands (e.g. sdk-core:test).

marshall-bean (changes requested):
A few small things.

projects/examples/package.json
0xalecks: this is a convention used by other web3 projects, like uniswap, so went with it

projects/examples/package.json
0xalecks: yeah, we can once we add tests

projects/sdk-core/src/constants/abi/ERC20.json
0xalecks: yes, we can commit the abis, these are pretty static

marshall-bean (approved):

Issue #308, opened 2023-02-01
Long error messages in popup


silochad:
**Describe the bug**

There's a function parse errors returned by ethers, but sometimes it fails and returns raw transaction calldata like this.

<img width="538" alt="image" src="https://user-images.githubusercontent.com/99763686/216128062-3232a861-cdb6-4104-9f93-283e89ffe8c6.png">

silochad:
@uncoolzero is currently looking into this

Issue #211, opened 2022-08-19
Improve generic error handling for forms


silochad:
## Description

We use the below pattern for alerting the user about errors when submitting a form. 

```ts
      } catch (err) {
        console.error(err);
        txToast ? txToast.error(err) : toast.error(parseError(err));
        formActions.setSubmitting(false);
      }
```

Problems:
- When `txToast` is used, a custom component is rendered with an infinite duration and a close button.
- `toast.error` only displays a string (without duration or close button).
- Error messages are often very long and contain raw JSON. Sometimes the error is so tall it overflows the screen, preventing it from being manually closed by a user.

Scope:
1. Find a better solution for parsing error messages from ethers to eliminate JSON and focus on the important part of the error.
2. Always use a component with controllable duration and a close button to give users maximum control.

silochad:
Error message parsing being tackled in in #308

PR #338, opened 2023-02-11
Improved error handling

uncoolzero:
More robust error handling, should prevent the tall error toast from ever appearing
Enabled the ability to copy the full error string to clipboard if we can't find a proper reason for the error

Addresses #308 and #211


silochad:
Looks like there's a TS-related error, could you run `yarn build` locally and debug? Here's what I see in the build logs:

```
3:41:30 PM: src/components/Barn/Actions/Buy.tsx(420,50): error TS2345: Argument of type 'Error' is not assignable to parameter of type 'Message'.
```

silochad (approved):

PR #343, opened 2023-02-13
SDK: Replace TokenValue and Tokens with version from sdk-core

0xalecks:


projects/sdk/src/lib/tokens.ts
```diff
@@ -46,120 +46,168 @@ export class Tokens {
     this.map = new Map();
 
     ////////// Ethereum //////////
+    const cid = this.sdk.chainId;
```
marshall-bean, line 49: We have plenty of storage :trollface: we can afford whole words (again, not blocking)

As in, `chainId` and `providerOrSigner`. We shouldn't have to stop and think about abbreviations when reading code.

marshall-bean (approved):

projects/sdk-wells/src/lib/Well.ts
```diff
@@ -45,28 +56,46 @@ export class Well {
...
+   *
+   * loadWell() -- loads everything
+   * loadWell({tokens: true}) - only loads tokens
+   *
+   */
+  async loadWell(options?: LoadOptions): Promise<void> {
+    // TODO: use a multicall
+    const toLoad = [];
+
+    if (!options) {
```
marshall-bean, line 73: I really think this is overkill for now. IMO we should avoid stuff like this. Not blocking/required change, just a comment about our process.

projects/sdk-wells/src/lib/Well.ts
```diff
@@ -45,28 +56,46 @@ export class Well {
...
+      toLoad.push(this.getName(), this.getLPToken(), this.getWell());
+    } else {
+      // console.log("load", options);
+      if (options.name) toLoad.push(this.getName());
+      if (options.lpToken) toLoad.push(this.getLPToken());
+      if (options.tokens || options.wellFunction || options.pumps || options.auger) toLoad.push(this.getWell());
+    }
+
+    await Promise.allSettled(toLoad);
```
marshall-bean, line 82: Do we want it to fail silently? I think we'd prefer `Promise.all` instead.

projects/sdk-wells/src/lib/Well.ts
```diff
@@ -45,28 +56,46 @@ export class Well {
...
+   *
+   */
+  async loadWell(options?: LoadOptions): Promise<void> {
+    // TODO: use a multicall
+    const toLoad = [];
+
+    if (!options) {
+      toLoad.push(this.getName(), this.getLPToken(), this.getWell());
+    } else {
+      // console.log("load", options);
```
marshall-bean, line 76: Errant `console.log`

marshall-bean (changes requested):

marshall-bean (approved):

PR #346, opened 2023-02-15
Wells USD values

cujowolf:
Update handlers to calculate and save well amounts in USD. Current limitation on implementation is an expectation of a two token well.

### DIFF
--- a/projects/subgraph-wells/src/WellHandler.ts
+++ b/projects/subgraph-wells/src/WellHandler.ts
@@ -21,7 +21,7 @@ export function handleAddLiquidity(event: AddLiquidity): void {
 
   checkForSnapshot(event.address, event.block.timestamp, event.block.number);
 
-  updateWellTokenBalances(event.address, event.params.tokenAmountsIn);
+  updateWellTokenBalances(event.address, event.params.tokenAmountsIn, event.block.timestamp, event.block.number);
 
   updateWellLiquidityTokenBalance(event.address, event.params.lpAmountOut);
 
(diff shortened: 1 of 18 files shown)


projects/subgraph-wells/subgraph.yaml
```diff
@@ -37,6 +39,10 @@ templates:
       abis:
         - name: Well
           file: ./abis/Well.json
+        - name: CurvePrice
+          file: ./abis/CurvePrice.json
+        - name: ERC20
+          file: ./abis/ERC20.json
       eventHandlers:
         - event: AddLiquidity(uint256[],uint256)
           handler: handleAddDeposit
```
marshall-bean, line 48: Shouldn't this be `handleAddLiquidity` ?

marshall-bean (changes requested):
Some observations about the data in the subgraph:
- lastPriceUSD showing as 0 for all tokens, leads all USD values to be 0 in the subgraph. Expected?
- swaps don't seem to be getting indexed, not yet sure why

projects/subgraph-wells/subgraph.yaml
cujowolf: Updated

marshall-bean (approved):

projects/subgraph-wells/src/WellHandler.ts
```diff
@@ -3,88 +3,95 @@ import { loadOrCreateAccount } from "./utils/Account";
...
+  incrementWellSwap,
+  incrementWellWithdraw,
+  loadWell,
+  updateWellLiquidityTokenBalance,
+  updateWellTokenBalances,
+  updateWellVolumes
+} from "./utils/Well";
 
 export function handleAddLiquidity(event: AddLiquidity): void {
+  loadOrCreateAccount(event.transaction.from);
```
marshall-bean, line 18: withdraws and deposits showing 0 for `amountUSD` always (not related to this line just mentioning it via PR review)

marshall-bean (changes requested):

marshall-bean (approved):

PR #353, opened 2023-02-21
Grown stalk per bdv

Brean0:
# Problem

Currently there are 2 problems in Beanstalk related to Deposits

1. Farmers with older Deposits may forfeit Stalk when converting from a Whitelisted Silo token with a higher `seedsPerBdv` to a Whitelisted Silo token with a lower `seedsPerBdv`
2.  `seedsPerBdv` must be static over time for each Whitelisted Silo token
3. `seedsPerBdv` cannot have Decimals

https://beanstalk-farms.notion.site/RFC-Cumulative-Stalk-per-BDV-2ed64ebd9aa6448f92e015c7a254b5c6

### DIFF
--- a/protocol/contracts/C.sol
+++ b/protocol/contracts/C.sol
@@ -51,7 +51,7 @@ library C {
     uint256 private constant SEEDS_PER_BEAN = 2;
     uint256 private constant STALK_PER_BEAN = 10000;
     uint256 private constant ROOTS_BASE = 1e12;
-    uint32 private constant SILOV3_START_SEASON = 10;
+    // uint32 private constant SILOV3_START_SEASON = 10; //maybe move to constant here eventually
 
 
     // Exploit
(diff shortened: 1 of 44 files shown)


Brean0:
@pizzaman1337  comments on here https://github.com/BeanstalkFarms/Beanstalk/pull/352

pizzaman1337:
todo: write tests that verify each token can be mowed in the same season/block (actually still need to make a multi-mow function)

Brean0:
in balanceOfGrownStalk, we call libSilo.stalkReward(), which calls

s.a[account].mowStatuses[token].lastCumulativeGrownStalkPerBdv
LibTokenSilo.cumulativeGrownStalkPerBdv(IERC20(token)), //get latest grown stalk per bdv for this token
s.a[account].mowStatuses[token].bdv
given that in this is often called where this is already read, think we can save gas by having an internal function to pass in the values

Brean0:
wrote solution to above in #354

Issue #189, opened 2022-12-28
Navigating straight to a listing/order fails to rehydrate position data


silochad:
Before refresh and after refresh:

<img width="1127" alt="image" src="https://user-images.githubusercontent.com/99763686/209745418-8bd00508-92e6-4a3c-9856-aecaf9a4e309.png">
<img width="1131" alt="image" src="https://user-images.githubusercontent.com/99763686/209745442-8143f518-0a5f-454a-8257-d660230b501f.png">

The graph is forgetting the pixel position of the selected listing/order during refresh, so believe it's drawing the lines & annotation at (0, 0).

tansanDOTeth:
I was inspecting this bug and found more issues with this particular graph. 

1. Select a point
2. Resize the window vertically
3. You'll see that the horizontal and vertical line hasn't been redrawn to the new scale.

![image](https://user-images.githubusercontent.com/92072158/212537191-c89caab7-54af-48cd-a54f-379f52cf3238.png)

tansanDOTeth:
I don't have the project setup, but looks like the bug is around here. The cursorPoint is memoized, so it's not being updated when the `selectedPoint` hasn't changed during the window resize. I am sure the charting system has an onChange callback that can be used to set the new `selectedPoint` and similarly on first load so it can trigger that memoization.

https://github.com/BeanstalkFarms/Beanstalk/blob/master/projects/ui/src/components/Market/PodsV2/Modules/MarketGraph.tsx#L268

tansanDOTeth:
@silochad Could you assign this to me please?

PR #355, opened 2023-02-23
[ISSUE-189] UI Fixes and general code improvements

tansanDOTeth:
This fixes issues https://github.com/BeanstalkFarms/Beanstalk/issues/189.

It seems this file had a lot of prettier formatting changes that was not applied. I went ahead and applied it in a separate commit first in b45bf92141da51c0014e113dc4b21cda5be11c2f, so you can easily review the bug fixes. Moreover, you'll find a refactor commit after that in 6439fae3f2c2499b6dd9de39d93477265d512cba, which shows how I made the code more DRY which should help with readability/comprehension. Afterwards, you'll find the commits for the actual fix. I would advise going through each commit individually since I segmented each change with the reviewer in mind.

### DIFF
--- a/projects/ui/src/components/Market/PodsV2/Modules/MarketGraph.tsx
+++ b/projects/ui/src/components/Market/PodsV2/Modules/MarketGraph.tsx
@@ -585,12 +585,17 @@ const Graph: FC<GraphProps> = ({
       }
     } else if (params.listingID) {
       const index = listings.findIndex((l) => l.id === params.listingID);
-      if (!selectedPoint || selectedPoint.index !== index) {
-        setSelectedPoint({
-          type: 'listing',
-          index,
-          coordinate: listingPositions[index],
-        });
+      const newSelectdPoint: SelectedPoint = {
+        type: 'listing',
+        index,
+        coordinate: listingPositions[index],
+      };
+      const hasPointChanged = selectedPoint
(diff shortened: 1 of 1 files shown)


0xalecks (changes requested):
Overall, PR is great, only issue is that we don't use Prettier for the UI. The rest of the monorepo does, just the UI hasn't been migrated over yet, it still uses eslint for formatting. For some reason, it's disabled at the top of the file.

Would you mind removing the prettier formatting? If you want to format, do it through eslint and feel free to remove the disabling, think it just needs a dependency array fixed. 

🙏 Thanks for the PR!

tansanDOTeth:
> Overall, PR is great, only issue is that we don't use Prettier for the UI. The rest of the monorepo does, just the UI hasn't been migrated over yet, it still uses eslint for formatting. For some reason, it's disabled at the top of the file.
> 
> Would you mind removing the prettier formatting? If you want to format, do it through eslint and feel free to remove the disabling, think it just needs a dependency array fixed.
> 
> 🙏 Thanks for the PR!

I can remove it, but there is rc file here in the ui directory here: https://github.com/BeanstalkFarms/Beanstalk/blob/master/projects/ui/.prettierrc

If you guys aren't using prettier, then I would suggest removing that as well. Alternatively, we can keep it and start using it moving forward. What would you suggest?

0xalecks (approved):

PR #356, opened 2023-02-23
(Claim + Plant) and do x

Bean-Sama:
### Overview

This PR introduces the ability for a farmer to perform a set of transactions along with any arbitrary contract call for efficiency (e.g. Claim Beans + Harvest Pods => Deposit claimed + harvested into the Silo). 

In addition, another goal was to start migrating the UI to use the SDK. Currently, only the workflow portion has migrated to use the new types, and there are many places that need to be refactored for there to be a seamless integration between the UI and the SDK. For example, the Token types differ between the UI and SDK, and the SDK uses BN.js whereas the UI uses BigNumber.js. Some convenience functions were created to help facilitate the integration of the SDK into the UI; however, we can make the rest of the refactors as we go.

### FormTxn and FormTxnBuilder
FormTxn is an enum that refers to individual contract calls can be called either before or after a primary set of instructions. Currently, `FormTxn` enumerates Mow, Plant, Enroot, Claim (Claim Silo Deposits), Harvest, and Rinse. `FormTxn`s are passed into `FormTxnBuilder`'s `compile` function along with a main set of instructions to be performed, upon which the actions are deduplicated, ordered, and built. 

```
compile(
    sdk: BeanstalkSDK,
    data: FormTxnBuilderInterface,
    getGenerators: (action: FormTxn) => StepGenerator[],
    operation: FarmWorkflow | StepGenerator[],
    amountIn: TokenValue | undefined,
    slippage: number,
    finalSteps?: {
      steps: StepGenerator[];
      overrideAmount?: TokenValue;
      additionalAmount?: TokenValue;
    }[]
  ) => {....}
```

'compile' knows how to order the FormTxns via a configuration object of type `FormTxnBuilderInterface`

```
export type FormTxnBuilderInterface = {
  preset: keyof typeof FormTxnBuilderPresets;
  primary: FormTxn[] | undefined;
  secondary: FormTxn[] | undefined;
  implied?: FormTxn[];
  exclude?: FormTxn[];
  transferToMode?: FarmToMode;
};
```

`primary` refers to the actions that must be performed before the main set of instructions (e.g., harvest pods -> Buy Fertilizer), and `secondary` refers to any actions that have no effect on the main set of instructions and can therefore be ordered in any way (e.g., Mow + Buy Fertilizer). `implied` refers to the actions that are already performed on the contract side when the main set of instructions is executed and therefore can be filtered out. `exclude` refers to the set of actions to exclude from the farm call.

`compile` must also be able to obtain the function call data for each FormTxn. It gets this data via calling the `getGenerators` parameter. In the current implementation, this is provided via a new hook `useFarmerFormTxns`. More on this below. 

In the case where any additional transactions need to be performed, `compile` takes an optional array of `finalSteps`. If defined, these are added to the end of the farm call. 

In the case where a final step requires a modified amount, a local-only step with the new amount is added prior to adding the steps to the workflow. This is accomplished via `FormTxnBuilder.getLocalOnlyStep`. It receives an optional 'additionalAmount'  and 'overrideAmount', which specify whether to add to the current amount returned from the previous steps or to completely override it with a new value. `getLocalOnlyStep` is used in other places as well to ensure the correct amounts are added to sections of a workflow.

#### useFarmerFormTxns
Gathers and packages the data needed to create a workflow step for each FormTxn. It additional provides a refetch function that handles refetching the necessary data prior to executing a farm call. 

#### useFamerFormTxnsSummary
Provides a summary of all FormTxns for the user. Determines whether or not the action is enabled for the user as well as any data the UI needs to show (e.g., the amount of claimable beans the user has, tooltips, etc. )

#### useFarmerFormTxnBalances
listens to Formik's form state changes and returns the balances a user is claiming / planting. 

#### useFarmerFormTxnActions
listens to Formik's form state changes and returns the actions the user is performing. The value returned form this hook is added as a prop to the TxnSummary. 

#### useFarmerDepositCrateFromPlant 
returns the deposit crate and amount that will be generated from planting. 

#### useDeepCompareMemoize
- used to run a deep comparison of the object values in a react dependency list and mutates only if the values are different. 
#### useDeepCompareMemo
- Uses useDeepCompareMemoize to do exactly what useMemo does but ignores new object references. Useful when new object references cause unnecessary re-renders.
#### useDeepCompareEffect
- accomplishes the same thing as useDeepCompareMemo but for react's useEffect.
#### useAsyncMemo
- a convenience hook that returns a memoized return value of an async function. 

### New components
- AddPlantTxnToggle: component used to add Plant to the form state
- AdditionalTxnsAccordion: component used to add FormTxns to the form state's 'secondary' actions.
- ClaimBeanDrawerContent: The drawer content for forms that enable the user to add FormTxns that claim beans to the form state's 'primary' actions.
- ClaimBeanDrawerToggle: Opens the ClaimBeanDrawer. Additionally displays which FormTxns have been selected.
- FormTxnOptionCard: Selectable card that handles displaying the information for a given specific FormTxn. 
- SelectionItem: Button component that handles a selected / hovered state. 
- TokenSelectionCard: Wrapper around SelectionItem to display a token and amount. 
- SelectionAccordion: MUI Accordion that handles the selection state of it's children. This component is used by 'AdditionalTxnsAccordion'. 
- Warning Alert: Generic MUI alert. Added for convenience. 

### additional components / hooks
These are temporary components that will replace the component/hook it replicated but required due to incompatible types between the UI and SDK. Eventually we will delete the old and rename these component/hook(s) it replicated. 
- UseQuoteWithParams: replica of useQuote but generic and receives additional parameters required via the function that calculates the 'amountOut' & workflow steps for a given set of instructions. 
- QuoteHandlerWithParams: replica of QuoteHandler but uses useQuoteWithParams instead of useQuote. 
- TokenSelectDialogNew: replica of TokenSelectDialog but supports the new SDK token types.


projects/ui/src/util/ClaimPlant.ts
```diff
@@ -0,0 +1,395 @@
...
+    })));
+  
+    const execute = () => farm.execute(amountIn, options);
+  
+    return { 
+      estimate, 
+      execute, 
+      actionsPerformed
+    };
+  }
```
Bean-Sama, line 369: @silochad Here is where the workflow is being built

projects/ui/src/hooks/farmer/claim-plant/useFarmerClaimPlantActions.ts
```diff
@@ -0,0 +1,195 @@
...
+        return claim(sdk, { crates, ...params });
+      },
+      [ClaimPlantAction.HARVEST]: (params) => {
+        const harvest = ClaimPlant.getAction(ClaimPlantAction.HARVEST);
+        return harvest(sdk, { plotIds, ...params });
+      },
+      [ClaimPlantAction.RINSE]: (params) => {
+        const rinse = ClaimPlant.getAction(ClaimPlantAction.RINSE);
+        return rinse(sdk, { tokenIds: fertilizerIds, ...params });
+      },
```
Bean-Sama, line 132: This might be helpful as well

Bean-Sama:
Temporarily updating to 'Ready for review' for deploy preview

projects/ui/src/hooks/chain/useTokenMap.ts
```diff
@@ -14,7 +15,9 @@ export default function useTokenMap<T extends Token>(
         // simply return the token. Otherwise we get the appropriate chain-
         // specific Token. This also dedupes tokens by address.
         const token = getChainToken(curr);
-        if (token) acc[token.address] = token;
+        // in the sdk, address of ETH is "". We need to use "eth" as key
+        const key = token instanceof Token && token.symbol === 'ETH' ? 'eth' : token.address;
+        if (token) acc[key] = token;
```
tansanDOTeth, line 20: This is more of a nitpick of mine than anything. Would this be better encapsulated in a function? Also, it would only evaluate the key if the token is truthy.

Ex,

```
// in the sdk, address of ETH is "". We need to use "eth" as key
const getTokenIndex = () => {
  if (token instanceof Token && token.symbol === 'ETH') {
    return 'eth';
  }

  return token.address;
}
...


...
if (token) acc[getTokenIndex(token)] = token;

```

projects/ui/src/hooks/farmer/form-txn/useFarmerFormTxnActions.ts
```diff
@@ -0,0 +1,131 @@
...
+import { ZERO_BN } from '~/constants';
+import useAccount from '~/hooks/ledger/useAccount';
+
+const isClaimingBeansAction = (action: FormTxn) => {
+  const isClaiming =
+    action === FormTxn.CLAIM ||
+    action === FormTxn.HARVEST ||
+    action === FormTxn.RINSE;
+
+  return isClaiming;
```
tansanDOTeth, line 19: What do you think about just returning this?
```

return  action === FormTxn.CLAIM ||
    action === FormTxn.HARVEST ||
    action === FormTxn.RINSE;
```

Or  use the arrow function without the curly braces

tansanDOTeth (changes requested):
I am not sure if your new file changes are being ported from some where, so apologies in advanced if that is the case. I may review again later, but the one thing I wanted to get clarity on is the `IS_DEV_ENV` bit. The var name suggests the intention should be a dev environment, which would be true when it's a production environment.

projects/sdk/src/lib/farm/actions/TransferDeposits.ts
```diff
@@ -0,0 +1,51 @@
...
+          amounts: this._amounts
+        });
+        return {
+          target: TransferDeposits.sdk.contracts.beanstalk.address,
+          callData: TransferDeposits.sdk.contracts.beanstalk.interface.encodeFunctionData("transferDeposits", [
+            this._signer, //
+            this._to, //
+            this._tokenIn, //
+            this._seasons, //
+            this._amounts, //
```
tansanDOTeth, line 43: Could you clean up the `//` here please?

projects/sdk/src/lib/farm/actions/WithdrawDeposit.ts
```diff
@@ -0,0 +1,39 @@
...
+          tokenIn: this._tokenIn,
+          seasons: this._season,
+          amounts: this._amount
+        });
+        return {
+          target: WithdrawDeposit.sdk.contracts.beanstalk.address,
+          callData: WithdrawDeposit.sdk.contracts.beanstalk.interface.encodeFunctionData("withdrawDeposit", [
+            this._tokenIn, //
+            this._season, //
+            this._amount //
```
tansanDOTeth, line 31: Could you also clean up the `//` here as well?

projects/ui/src/components/Barn/Actions/Buy.tsx
```diff
@@ -173,266 +257,377 @@ const BuyForm : FC<
...
-const Buy : FC<{}> = () => {
-  // Wallet connection
-  const account = useAccount();
-  const { data: signer } = useSigner();
-  const provider = useProvider();
-  const chainId = useChainId();
-  const fertContract = useFertilizerContract(signer);
-  const beanstalk = useBeanstalkContract(signer);
+const Buy: FC<{}> = () => {
+  ///
```
tansanDOTeth, line 282: Could you remove this please?

projects/ui/src/hooks/farmer/useFertilizerSummary.ts
```diff
@@ -14,35 +13,146 @@ import { USDC } from '~/constants/tokens';
...
+  //     }
+      
+  //     return memo;
+  //   }, {
+  //     usdc:     new BigNumber(0),  // The amount of USDC to be swapped for FERT.
+  //     fert:     new BigNumber(0),
+  //     humidity: humidity,          //
+  //     actions:  [] as Action[],    // 
+  //   })
+  // }, [humidity, tokens, eth, usdc]);
```
tansanDOTeth, line 67: Is there a reason for preserving this? If not, could we delete it? Same question for other commented out code in this file

projects/ui/src/hooks/beanstalk/useTemperature.ts
```diff
@@ -0,0 +1,10 @@
+import { useSelector } from 'react-redux';
+import { AppState } from '~/state';
+
+export default function useTemperature() {
+  const weather = useSelector<
+    AppState,
+    AppState['_beanstalk']['field']['weather']['yield']
+  >((state) => state._beanstalk.field.weather.yield);
+  return weather;
```
tansanDOTeth, line 9: Could you just return the useSelector here?

```
return useSelector<
    AppState,
    AppState['_beanstalk']['field']['weather']['yield']
  >((state) => state._beanstalk.field.weather.yield);
```

projects/ui/src/components/Common/Form/TokenSelectDialogNew.tsx
```diff
@@ -0,0 +1,301 @@
...
+                            <HelpOutlineIcon
+                              sx={{
+                                color: 'text.secondary',
+                                display: 'inline',
+                                mb: 0.5,
+                                fontSize: '12px',
+                              }}
+                            />
+                          </Tooltip>
+                        ) : null} */}
```
0xalecks, line 260: comments

projects/ui/src/components/Silo/Actions/Withdraw.tsx
```diff
@@ -364,3 +397,64 @@ const Withdraw : FC<{ token: ERC20Token; }> = ({ token }) => {
...
+            transition: `height ${CONFIRM_DELAY}ms linear`,
+            height: '100%',
+          }
+        }}
+        className={fill}
+      />
+      Confirm Withdrawal
+    </Button>
+  </StyledDialogActions>
+</StyledDialog> */
```
0xalecks, line 460: Remove these

0xalecks (changes requested):

projects/ui/src/hooks/farmer/form-txn/useFarmerFormTxnsSummary.ts
```diff
@@ -0,0 +1,337 @@
+import { useCallback, useMemo } from 'react';
+import { FarmToMode, Token } from '@beanstalk/sdk';
+import BigNumber from 'bignumber.js';
+import { ZERO_BN } from '~/constants';
+import useSdk, { getNewToOldToken } from '../../sdk';
+
+import useFarmerFertilizer from '../useFarmerFertilizer';
+import useFarmerField from '../useFarmerField';
+import useFarmerSilo from '../useFarmerSilo';
+import useRevitalized from '../useRevitalized';
```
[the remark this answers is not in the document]
Bean-Sama: can do!

projects/ui/src/util/BigNumber.ts
```diff
@@ -11,3 +13,22 @@ export const BN = (v: BigNumber.Value) => new BigNumber(v);
...
 // };
+
+export function normalizeBN(
+  value: BigNumber | undefined | null,
+  _gt?: BigNumber.Value,
+) {
+  return (value && value.gt(_gt || 0) ? value : ZERO_BN);
+}
+
+export function tokenValueToBN(
```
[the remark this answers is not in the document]
Bean-Sama: So the reason why I've added these is because we use a different BigNumber package in the sdk. I believe in the sdk we use bn.js and in the ui we use bignumber.js.

projects/ui/src/components/Silo/Actions/Convert.tsx
```diff
@@ -364,210 +413,312 @@ const ConvertForm : FC<
...
+          return prev;
+        },
+        [null, null] as [Token | null, Token | null]
+      );
+
+      if (!inToken || !outToken) throw new Error('conversion unavailable');
+
+      const amountIn = inToken.amount(_amountIn.toString());
+      const { deposited } = await sdk.silo.getBalance(inToken, account, {
+        source: DataSource.LEDGER,
```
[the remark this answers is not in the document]
Bean-Sama: This is true, but the only reason why I couldn't use sdk.silo.siloConvert.convertEstimate is because the function checks the balance of the user's silo deposits and will fail if I pass in an additional crate for what will be converted via planting before

projects/sdk/src/lib/farm/actions/ClaimWithdrawal.ts
```diff
@@ -0,0 +1,40 @@
...
+          tokenIn: this._tokenIn,
+          seasons: this._season,
+          to: this._to
+        });
+        return {
+          target: ClaimWithdrawal.sdk.contracts.beanstalk.address,
+          callData: ClaimWithdrawal.sdk.contracts.beanstalk.interface.encodeFunctionData("claimWithdrawal", [
+            this._tokenIn, //
+            this._season, //
+            this._to
```
tansanDOTeth, line 32: Could you remove `//` please?

tansanDOTeth (changes requested):
I don't think `hooks/display` is the correct directory for these types of hooks. It could sit in `hooks/app` or just in the index of `hooks`. 

image

projects/ui/src/util/FormTxns.ts
tansanDOTeth, line 1: This file can benefit greatly from more OOP design. Ex, strategy pattern or command pattern or DI to help with more encapsulation and inversion of control. Better design will help with readability, review, and maintenance.

Here is a general overarching design that should be used in this case. All commands should be in a separate directory.

```javascript

class HarvestCommand {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class RinseCommand {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class ClaimCommand {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class MowCommand {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class PlantCommand {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}


class EnrootCommand {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class FormTxnBuilder {
  private static harvest: FormTxnFunction<FormTxn.HARVEST> = HarvestCommand

  private static rinse: FormTxnFunction<FormTxn.RINSE> = RinseCommand

  private static claim: FormTxnFunction<FormTxn.CLAIM> = ClaimCommand

  private static mow: FormTxnFunction<FormTxn.MOW> = MowCommand

  private static plant: FormTxnFunction<FormTxn.PLANT> = PlantCommand

  private static enroot: FormTxnFunction<FormTxn.ENROOT> = EnrootCommand
  };
```

cc: @0xalecks for thoughts and will post in Discord

projects/ui/src/util/FormTxns.ts
```diff
@@ -0,0 +1,587 @@
...
+  };
+
+  private static functionMap: { [key in FormTxn]: FormTxnFunction<key> } = {
+    [FormTxn.MOW]: FormTxnBuilder.mow,
+    [FormTxn.PLANT]: FormTxnBuilder.plant,
+    [FormTxn.ENROOT]: FormTxnBuilder.enroot,
+    [FormTxn.RINSE]: FormTxnBuilder.rinse,
+    [FormTxn.HARVEST]: FormTxnBuilder.harvest,
+    [FormTxn.CLAIM]: FormTxnBuilder.claim,
+  };
```
tansanDOTeth, line 376: Okay, after reviewing how this is used. Here is an updated suggestion.

This can be pulled out into a factory into it's own file.

Ex, 

```javascript
// Some strategy folder 
class HarvestStepStrategy {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class RinseStepStrategy {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class ClaimStepStrategy {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class MowStepStrategy {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class PlantStepStrategy {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}


class EnrootStepStrategy {
  constructor(sdk, params) {
    ...
  }
  getSteps() {
    ...
  }
  estimateGas() {
    ...
  }
}

class FormTxnStepFactory {
  getStep(step) {
    switch(step) {
      case FormTxn.MOW:{ 
        return MowStepStrategy
      }
      case FormTxn.PLANT:{ 
        return PlantStepStrategy
      }
      case FormTxn.ENROOT:{ 
        return EnrootStepStrategy
      }
      case FormTxn.RINSE:{ 
        return RinseStepStrategy
      }
      case FormTxn.HARVEST:{ 
        return HarvestStepStrategy
      }
      case FormTxn.CLAIM:{ 
        return ClaimStepStrategy
      }
      default:
        throw new Error('Step does not exist')
    }
  }
}
```

If you have the above code, you can remove the following code: 
* https://github.com/BeanstalkFarms/Beanstalk/blob/179fa41fa2123ee954718c52d62d843c96e6a597/projects/ui/src/util/FormTxns.ts#L369-L376
* https://github.com/BeanstalkFarms/Beanstalk/blob/179fa41fa2123ee954718c52d62d843c96e6a597/projects/ui/src/util/FormTxns.ts#L385-L387
* https://github.com/BeanstalkFarms/Beanstalk/blob/179fa41fa2123ee954718c52d62d843c96e6a597/projects/ui/src/util/FormTxns.ts#L117-L358

Then you can refactor all calls to `FormTxnBuilder.getFunction`. 

Here is an example of how it can be refactored:
https://github.com/BeanstalkFarms/Beanstalk/blob/179fa41fa2123ee954718c52d62d843c96e6a597/projects/ui/src/components/Barn/Actions/Rinse.tsx#L214-L219

```javascript
// Call Exmaple
const params = {
  tokenIds: farmerBarn.balances.map((bal) => bal.token.id.toString()),
  toMode: values.destination,
};
const stepKlass = FormTxnStepFactory.getStep(FormTxn.RINSE);
const rinse = new stepKlass(sdk, params);
const rinseStep = rinse.getSteps();
```

projects/ui/src/util/BigNumber.ts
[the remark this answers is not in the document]
0xalecks: Oh, you're right, missed that, my bad

PR #357, opened 2023-02-23
Pull out common subgraph components

cujowolf:
This pulls out common items used by the subgraphs such as contract ABIs and constants, decimal functions


projects/subgraph-beanstalk/src/FarmHandler.ts
```diff
@@ -1,38 +1,35 @@
 import { Address, BigInt } from "@graphprotocol/graph-ts";
 import { InternalBalanceChanged } from "../generated/Farm/Beanstalk";
 import { loadBeanstalk } from "./utils/Beanstalk";
-import { BEANSTALK } from "./utils/Constants";
+import { BEANSTALK } from "../../subgraph-core/utils/Constants";
```
marshall-bean, line 4: I think we should make the core a proper package, and these references should use the package name / reference instead, such as `@beanstalk/subgraph-core` with BEANSTALK constants exported.

Example: https://github.com/BeanstalkFarms/Beanstalk/blob/b3222275b882bcb827f8675c2251e7a7dc3f7830/projects/sdk-wells/src/lib/Well.ts#L1

But let's discuss in the discord.

marshall-bean (changes requested):

marshall-bean (approved):

Issue #361, opened 2023-02-28
Design nits for forms


silochad:
This is a running issue which we can update throughout Q1/Q2 with design nitpicks and fixes.

### All forms

**(1)**
![Feb-28-2023 11-05-28](https://user-images.githubusercontent.com/99763686/221925462-81d1bb7d-1ac9-4012-a933-93a4c3f509db.gif)

- [ ] (1a) Input before hover: Two different grey borders for inputs
- [ ] (1b) Input on focus: New input has green outline, old has blue
- [ ] (1c) Button before hover: 
  - The one inside the input has grey border
  - The one outside the input has a light blue border
- [ ] (1d) Button on hover: 
  - The one inside the input receives a blue background
  - The one outside the input receives a green border with light-green background

**(2)**
![image](https://user-images.githubusercontent.com/99763686/221926467-da11e0bd-5287-4aa6-9331-fb9637b1fa45.png)

- [ ] (2a) The "Balance: 100 (Max)" bit should be shifted so that the right side is aligned with the button above
- [ ] (2b) Should the balance label say "Circulating Balance", etc.? In practice, I found it a bit hard to parse the icon at a glance.

**(3)**
![image](https://user-images.githubusercontent.com/99763686/221926820-9ba8e869-3878-481a-8bc4-35fe05b887c4.png)

- [ ] (3a) Button hover: Border is blue before, green after. Confirm w/ design that this is the behavior desired.
- [ ] (3b) Some modals like this one have extra padding at the bottom. Remove the extra padding.
- [ ] (3c) Add extra padding above the buttons but below the header, so that the space from the "Destination" to the top line of the modal is the same as the space from "Destination" to the first button. (We might need to make it more like 75% of the topline distance). Other modals:
  - [ ] Farmer's Almanac

**(4)**
![Feb-28-2023 11-18-26](https://user-images.githubusercontent.com/99763686/221928937-4b557b68-f703-45c3-ae18-9f2914029629.gif)
- [ ] (4a) Believe this is switching from white -> blue border, and the blue just happens to blend in with the header background.
  - [ ] CHECK WITH DESIGN: May be fixed by theme update in Q1.

**(5)**
![Feb-28-2023 11-20-56](https://user-images.githubusercontent.com/99763686/221929124-8d3dbf2b-3a0f-4c46-9a75-879baeca3a6b.gif)
- [ ] (5a) Nav button hover background isn't legible on the blue background.
[...]

PR #367, opened 2023-03-08
Design Tweaks

uncoolzero:
Done as of 26/03, this looks to address some of the issues raised in #361 

(1)
- [x] (1a) Input before hover: Two different grey borders for inputs
- [x] (1b) Input on focus: New input has green outline, old has blue
- [x] (1c) Button before hover:
The one inside the input has grey border
The one outside the input has a light blue border
- [x] (1d) Button on hover:
The one inside the input receives a blue background
The one outside the input receives a green border with light-green background

(2)

- [x] (2a) The "Balance: 100 (Max)" bit should be shifted so that the right side is aligned with the button above
- [x] (2b) Should the balance label say "Circulating Balance", etc.? In practice, I found it a bit hard to parse the icon at a glance.

(3)

- [x] (3a) Button hover: Border is blue before, green after. Confirm w/ design that this is the behavior desired.
- [x] (3b) Some modals like this one have extra padding at the bottom. Remove the extra padding.
- [x] (3c) Add extra padding above the buttons but below the header, so that the space from the "Destination" to the top line of the modal is the same as the space from "Destination" to the first button. (We might need to make it more like 75% of the topline distance). Other modals:
  - [x] Farmer's Almanac

(4)

- [x] (4a) Believe this is switching from white -> blue border, and the blue just happens to blend in with the header background.
  - [x] CHECK WITH DESIGN: May be fixed by theme update in Q1.

(5)

- [x] (5a) Nav button hover background isn't legible on the blue background.
  - [x] CHECK WITH DESIGN: May be fixed by theme update in Q1.

(L1) (Low Priority)

- [ ] (L1a) Border radius matches in pixel terms here, but doesn't match visually because of the padding difference. The border radius on the button needs to be reduced. Forget the algo for calculating this but it's floating around somewhere.

(L2) (Low Priority)

- [ ] (L2a) The BeaNFT icon is hard to decipher and looks strange IMO. Should we use a real image?
- [x] (L2b) The icon colors are green while the theme is blue
- [x] (L2c) The arrow in the dropdown line item isn't aligned with the text
- [x] (L2d) Swap should probably be moved to a top-level nav icon or button since it's becoming increasingly important


projects/ui/src/components/Common/Form/BorderEffect.tsx
```diff
@@ -22,31 +22,23 @@ const BorderEffect: FC<
...
     };
 
     const handleClickAway = () => {
       if (!ref.current || !activeRef.current || disabled) return;
-      ref.current.style.padding = '1px';
-      ref.current.style.border = '1px solid';
-      ref.current.style.borderColor = BeanstalkPalette.inputGrey;
+      ref.current.style.boxShadow = `inset 0 0 0 1px ${BeanstalkPalette.inputGrey}`;
       activeRef.current = false;
     };
```
silochad, line 43: Future work: This pattern is sufficient for now but let's explore aways to avoid using direct DOM modification for this, CSS will be cleaner & faster.

projects/ui/src/components/Nav/FolderMenu.tsx
```diff
@@ -84,6 +101,25 @@ const FolderMenu: FC<{
...
+    function handleResize() {
+      if (window.innerWidth < theme.breakpoints.values.lg && !mobileWindow) {
+        setMobileWindow(true);
+      } else {
+        setMobileWindow(false);
+      }
+    }
+    window.addEventListener('resize', handleResize);     
+    return () => window.removeEventListener('resize', handleResize);
+  }, []);
```
silochad, line 114: Future work: Switch this to the `useMediaQuery` API which optimizes event listeners. This implementation works but will be slow, especially when the user is dragging their screen size to be smaller.

https://mui.com/material-ui/react-use-media-query/

Alternatively, the JS-based style strategy could be replaced with a CSS-based one where a new component is rendered after a certain breakpoint.

projects/ui/src/components/Nav/FolderMenu.tsx
```diff
@@ -139,11 +176,24 @@ const FolderMenu: FC<{
...
         sx={{
-          zIndex,
+          zIndex: zIndex,
+          visibility: mobileWindow ? 'hidden' : 'visible',
         }}
+        modifiers={[
+          {
+            name: 'computeStyles',
+            options: {
+              gpuAcceleration: false,
```
silochad, line 189: Fine to leave this, but curious for my own learning if there was any particular reason

silochad (approved):

projects/ui/src/components/Nav/FolderMenu.tsx
silochad: Tracking here: #438

PR #379, opened 2023-03-20
#Issue298 Remove hardcoded RPC URL  

nickkatsios:
Removed the hardcoded rpc url from the sdk.


projects/sdk/src/utils/TestUtils/BlockchainUtils.ts
```diff
@@ -80,7 +80,7 @@ export class BlockchainUtils {
     await this.sdk.provider.send("anvil_reset", [
       {
         forking: {
-          jsonRpcUrl: "https://eth-mainnet.g.alchemy.com/v2/f6piiDvMBMGRYvCOwLJFMD7cUjIvI1TP"
+          jsonRpcUrl: "PROVIDER_RPC_URL"
```
0xalecks, line 83: This ENV var doesn't exist anywhere else; this change would break things

0xalecks (changes requested):

projects/sdk/src/utils/TestUtils/BlockchainUtils.ts
nickkatsios: Would it be ok if i added a .env file on the TestUtils folder and got the value from process.env? So that everyone could use their own rpc url.

nickkatsios: Rephrase: Should i get the rpc url through process.env? So that everyone can create their own .env file and put any api keys or urls here.

nickkatsios: Changed. Should work now with existing .env-sample file.

PR #380, opened 2023-03-21
Zero Withdraw - Tweaks

Brean0:
**Todo**
- [ ] @Brean0 Itemized list of changes in this PR & their rationale; add to this GitHub comment
- [ ] @Brean0 Relay list of changes to Halborn in next meeting for review
- [ ] @Brean0 Natspec documentation and inline comments for everything in scope of Zero Withdraw
- [ ] @silochad Code review
- [ ] @silochad Documentation and naming review
- [ ] @pizzaman1337 Code review


protocol/contracts/beanstalk/silo/SiloFacet/SiloExit.sol
```diff
@@ -141,9 +141,6 @@ contract SiloExit is ReentrancyGuard {
     /**
      * @notice Returns the balance of Earned Beans for `account`. Earned Beans
      * are the Beans distributed to Stalkholders during {Sun-rewardToSilo}.
-     * @dev in the case where a user calls balanceOfEarned beans during the vesting period,
-     * we have to manually calculate the deltaRoots and newEarnedRoots, as they are not stored in the state.
-     * this is done in {_calcRoots} and {_balanceOfEarnedBeansVested}.
```
pizzaman1337, line 146: This seems like a useful comment - was it removed because it's no longer true?

protocol/contracts/beanstalk/silo/SiloFacet/TokenSilo.sol
```diff
@@ -94,6 +94,22 @@ contract TokenSilo is Silo {
         uint256 amount,
         uint256[] bdvs
     ); //add bdv[] here? in favor of array
+    
+    // per the zero withdraw update, there is no claiming function for withdraws.abi
+    // events are kept for backwards compatibility
```
pizzaman1337, line 99: I thought we decided there could be some in-flight withdrawals that have not yet been claimed, so those claim functions need to be kept accessible? or is this comment just referring to the exact abi file where those will be visible?

pizzaman1337:
This may be out of the scope of this MR, but is there a public view function that returns a boolean on if we're currently in the "morning" for a particular sunrise? one that does a check like `if (block.number - s.season.sunriseBlock <= EARNED_BEAN_VESTING_BLOCKS) {` and returns true/false if in the morning

protocol/contracts/beanstalk/silo/SiloFacet/SiloFacet.sol
```diff
@@ -299,6 +299,10 @@ contract SiloFacet is TokenSilo {
         return _plant(msg.sender, token);
     }
 
+    function plantForAccount(address account) external payable returns (uint256 beans) {
+        return _plant(account);
+    }
```
pizzaman1337, line 304: I suppose the idea is to expose this so that anyone can plant for anyone else is for Tractor purposes? If you plant in the morning then can you still plant later within the same season? Just trying to think if there's something malicious you could do if you planted for someone else (like slip in a plant tx right before their <whatever> tx if you see someone is trying to do something with another tx that's visible in the mempool and planting for them could mess them up somehow)

PR #381, opened 2023-03-22
Stem re addclaim

Brean0:


protocol/contracts/beanstalk/silo/SiloFacet/MigrationFacet.sol
```diff
@@ -55,5 +48,28 @@ contract MigrationFacet is ReentrancyGuard {
         LibLegacyTokenSilo._migrateNoDeposits(account);
     }
 
+    //////////////////////// CLAIM ////////////////////////
+
+    // as of the zero withdraw update, these functions are no longer used. 
+    // However, these are kept for backwards compatability.
+    function claimWithdrawal(
```
silochad, line 55: I previously wrote natspec comments for both of these functions, take a look back to before they were removed and reintroduce them with the following changes:

- Replace the @notice line with `@notice DEPRECATED: ...`
- Add or extend the @dev comment explaining why these were deprecated.

When explaining deprecation, say something like:
> The Zero Withdraw update removed the two-step withdraw & claim process. These functions are left for backwards compatibility, to allow pending withdrawals from before the update to be claimed.

protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -186,7 +193,8 @@ contract ConvertFacet is ReentrancyGuard {
                 token,
                 stems[i],
                 amounts[i],
-                bdv
+                bdv,
+                true // converting "burns" the old ERC1155 deposit, and "mints" a new ERC1155 deposit.
```
silochad, line 197: Talk with @pizzaman1337 to see if this boolean would be better as an enum. I don't think it matters from a gas perspective (correct me if that's wrong) but would be way clearer from a code perspective.

protocol/contracts/libraries/Silo/LibLegacyTokenSilo.sol
```diff
@@ -505,4 +530,56 @@ library LibLegacyTokenSilo {
         }
         return 0;
     }
+
+    ////////////////////////// CLAIM ///////////////////////////////
+    // as of the zero withdraw update, these functions are no longer used. 
+    // However, these are kept for backwards compatability.
```
silochad, line 536: Leave one line break between this and the section header. 

For functions in a contract, always use the block comment and an `@dev` natspec comment

PR #382, opened 2023-03-23
Update wells abi and add deadline param

marshall-bean:
Finally discovered why my sample data scripts broke.


projects/sdk-wells/src/lib/Well.ts
```diff
@@ -235,20 +239,23 @@ export class Well {
     amountIn: TokenValue,
     minAmountOut: TokenValue,
     recipient: string,
+    _deadline: number | string,
```
0xalecks, line 242: Does the contract accept number | string?

0xalecks (changes requested):

projects/sdk-wells/src/lib/Well.ts
```diff
@@ -33,6 +33,8 @@ export type PreloadOptions = {
   reserves?: boolean;
 };
 
+const MAX_UINT256 = "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
```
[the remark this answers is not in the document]
marshall-bean: Thanks will change.

projects/sdk-wells/src/lib/Well.ts
marshall-bean: I should fix this, looking at it again it should just be a string.

0xalecks (approved):

PR #396, opened 2023-04-04
Create DEVELOPER.md

0xalecks:


DEVELOPER.md
```diff
@@ -0,0 +1,83 @@
...
+- Install Foundry with `curl -L https://foundry.paradigm.xyz | bash` and reopen your terminal
+- Run `foundryup` to ensure you have downloaded the latest version
+- Start a locally forked node with the following command:
+
+```bash
+# optionally add
+# --fork-block-number <BLOCK_NUMBER>
+# if you want a specific point in time
+
+anvil --fork-url <FORK_RPC>  --chain-id 1337 --block-time 12
```
marshall-bean, line 23: We should include a note about the subgraph.

E.g.

"If you are using anvil with the local subgraph node, include the `--disable-block-gas-limit` option to avoid issues when indexing.

marshall-bean (changes requested):

marshall-bean (approved):

PR #413, opened 2023-04-21
Remove liquidity balanced one token and custom

marshall-bean:
Remove liquidity:
- balanced (50/50 ratio)
- one token
- custom ratio

### DIFF
--- a/projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
+++ b/projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
@@ -1,0 +1,320 @@
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import { TokenInput } from "src/components/Swap/TokenInput";
+import { Token, TokenValue } from "@beanstalk/sdk";
+import { useAllTokensBalance } from "src/tokens/useTokenBalance";
+import styled from "styled-components";
+import { images } from "src/assets/images/tokens";
+import { useAccount } from "wagmi";
+import { ContractReceipt } from "ethers";
+import { Well } from "@beanstalk/sdk/Wells";
+import { useQuery } from "@tanstack/react-query";
+
+type RemoveLiquidityProps = {
+  well: Well;
+};
+
+enum REMOVE_LIQUIDITY_MODE {
+  Balanced,
+  OneToken,
+  Custom
+}
+
+type LiquidityAmounts = {
+  [key: number]: TokenValue;
+};
+
+export const RemoveLiquidity = ({ well }: RemoveLiquidityProps) => {
+  const { address } = useAccount();
+
+  const [wellTokens, setWellTokens] = useState<Token[]>();
+  const [wellLpToken, setWellLpToken] = useState<Token | null>(null);
+  const [lpTokenAmount, setLpTokenAmount] = useState<TokenValue | undefined>();
+  const [receipt, setReceipt] = useState<ContractReceipt | null>(null);
+  const [isLoadingAllBalances, setIsLoadingAllBalances] = useState(true);
+  const [removeLiquidityMode, setRemoveLiquidityMode] = useState<REMOVE_LIQUIDITY_MODE>(REMOVE_LIQUIDITY_MODE.Balanced);
+  const [singleTokenIndex, setSingleTokenIndex] = useState<number>(0);
+  const [amounts, setAmounts] = useState<LiquidityAmounts>({});
+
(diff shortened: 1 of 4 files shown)


projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
```diff
@@ -0,0 +1,320 @@
...
+  Balanced,
+  OneToken,
+  Custom
+}
+
+type LiquidityAmounts = {
+  [key: number]: TokenValue;
+};
+
+export const RemoveLiquidity = ({ well }: RemoveLiquidityProps) => {
```
0xalecks, line 26: I've been using this type for all react functional components, it adds some built in react things to the props, like `children`, tho not needed in this case:

```
import { FC } from "src/types";
export const RemoveLiquidity = FC<RemoveLiquidityProps>({ well }) => {
```

projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+      setWellTokens(tokens);
+      setAmounts(initialAmounts);
+    }
+  }, [well]);
+
+  useEffect(() => {
+    if (well.lpToken) {
+      let lpTokenWithMetadata = well.lpToken;
+      lpTokenWithMetadata.setMetadata({ logo: images[well.lpToken.symbol] ?? images.DEFAULT });
```
0xalecks, line 65: we should have a different default image for our Well LP tokens. cc @sweetredbeans

projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
```diff
@@ -0,0 +1,320 @@
...
+    data: customRatioQuote,
+    isLoading: loadingCustomRatioQuote,
+    isError: customRatioQuoteError
+  } = useQuery(["wells", address, removeLiquidityMode, amounts], async () => {
+    if (!atLeastOneAmountNonzero() || removeLiquidityMode !== REMOVE_LIQUIDITY_MODE.Custom) {
+      return null;
+    }
+
+    return well.removeLiquidityImbalancedQuote(Object.values(amounts));
+  });
```
0xalecks, line 105: should we put these in a single hook, like 
useRemoveLiquidityQuote(quoteType, amounts[], tokenIndex?), then based on the params, return the correct useQuery result?

Also, for the useQuery cache keys:
- not sure why we need `removeLIquidityMode`
- the amounts used for quote... if we really want to cache the quote result, we should change the default ttls

projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
```diff
@@ -0,0 +1,320 @@
...
+  const [amounts, setAmounts] = useState<LiquidityAmounts>({});
+
+  const { isLoading: isAllTokenLoading, refetch: refetchBalances } = useAllTokensBalance();
+
+  const atLeastOneAmountNonzero = useCallback(() => Object.values(amounts).filter((amount) => amount.value.gt("0")).length > 0, [amounts]);
+
+  useEffect(() => {
+    const fetching = isAllTokenLoading;
+    fetching ? setIsLoadingAllBalances(true) : setTimeout(() => setIsLoadingAllBalances(false), 500);
+  }, [isAllTokenLoading]);
```
0xalecks, line 45: can we skip this and let the parent, Wells page, handle it? Maybe not show this component if the tokens are still loading?

0xalecks (approved):

projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
marshall-bean: Good idea, I'll implement that in the next PR.

projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
marshall-bean: Yes they do belong in a separate hook for sure. I was treating it as a "dependency" as the docs suggested.

> if we really want to cache the quote result, we should change the default ttls
Can you elaborate on this one?

projects/dex-ui/src/components/Liquidity/RemoveLiquidity.tsx
marshall-bean: I think it's great as is. :trollface:

PR #418, opened 2023-04-26
Feedback from earlier PR

marshall-bean:
Feedback from PR https://github.com/BeanstalkFarms/Beanstalk/pull/413

### DIFF
--- a/projects/dex-ui/src/wells/useLiquidityQuote.tsx
+++ b/projects/dex-ui/src/wells/useLiquidityQuote.tsx
@@ -1,0 +1,82 @@
+import { Well } from "@beanstalk/sdk/Wells";
+import { useQuery } from "@tanstack/react-query";
+import { Token, TokenValue } from "@beanstalk/sdk";
+import { useCallback } from "react";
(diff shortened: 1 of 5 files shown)


projects/dex-ui/src/wells/useLiquidityQuote.tsx
```diff
@@ -0,0 +1,82 @@
...
+    oneToken: {
+      oneTokenQuote,
+      loadingOneTokenQuote,
+      oneTokenQuoteError
+    },
+    custom: {
+      customRatioQuote,
+      loadingCustomRatioQuote,
+      customRatioQuoteError
+    }
```
0xalecks, line 80: would be better to just return one single "useQuery" result, depending on the parameters. something like:
```

```

marshall-bean: I tried this and I think it's better to keep them separate like I have them. The reason is the types of the quotes are different depending on whether we are one token vs balanced vs custom.

0xalecks (approved):

PR #435, opened 2023-05-02
Update add liquidity to reflect design

marshall-bean:
Screenshot 2023-05-03 at 8 05 12 AM

### DIFF
--- a/projects/dex-ui/src/components/Liquidity/AddLiquidity.tsx
+++ b/projects/dex-ui/src/components/Liquidity/AddLiquidity.tsx
@@ -14,6 +14,7 @@ import useSdk from "src/utils/sdk/useSdk";
 import { formatEther } from "ethers/lib/utils.js";
 import useEthPrice from "./useEthPrice";
 import { ensureAllowance, hasMinimumAllowance } from "./allowance";
+import { Log } from "src/utils/logger";
 
 type AddLiquidityProps = {
   well: Well;
@@ -42,9 +43,9 @@ export const AddLiquidity = ({ well, txnCompleteCallback, slippage, slippageSett
     const _tokenAllowance = [];
     for (let [index, token] of well.tokens!.entries()) {
       // only check approval if this token has an amount gt zero
-      if (amounts[index] && amounts[index].value.gt("0")) {
+      if (amounts[index] && amounts[index].gt(0)) {
         const tokenHasMinAllowance = await hasMinimumAllowance(address, well.address, token, amounts[index].value.toBigNumber());
-        console.debug(`Token ${token.symbol} with amount ${amounts[index].toHuman()} has approval ${tokenHasMinAllowance}`);
+        Log.module("addliquidity").debug(`Token ${token.symbol} with amount ${amounts[index].toHuman()} has approval ${tokenHasMinAllowance}`);
         _tokenAllowance.push(tokenHasMinAllowance);
       } else {
         _tokenAllowance.push(false);
@@ -52,12 +53,6 @@ export const AddLiquidity = ({ well, txnCompleteCallback, slippage, slippageSett
     }
     setTokenAllowance(_tokenAllowance);
   }, [address, amounts, well.address, well.tokens]);
-
-  const allowanceCallback = useCallback(() => {
-    // TODO: Open notification??
-    console.log("allowance updated");
-    checkMinAllowanceForAllTokens();
-  }, [checkMinAllowanceForAllTokens]);
 
   // Once we have our first quote, we show the details.
   // Subsequent quote invocations shows a spinner in the Expected Output row
@@ -87,16 +82,13 @@ export const AddLiquidity = ({ well, txnCompleteCallback, slippage, slippageSett
     }
   }, [well.tokens]);
 
-  const {
-    data: quote
-    // isLoading: loadingQuote,
-    // isError: quoteError
-  } = useQuery(["wells", address, amounts], async () => {
+  const allTokensHaveMinAllowance = useMemo(() => tokenAllowance.filter((a) => a === false).length === 0, [tokenAllowance]);
+
+  const { data: quote } = useQuery(["wells", "quote", "addliquidity", address, amounts, allTokensHaveMinAllowance], async () => {
     if (!atLeastOneAmountNonzero()) {
       return null;
     }
 
-    // TODO: Scope
     if (!allTokensHaveMinAllowance) {
       return null;
     }
@@ -106,21 +98,22 @@ export const AddLiquidity = ({ well, txnCompleteCallback, slippage, slippageSett
     // so we show the quote details page on first quote
     setShowQuoteDetails(true);
 
-    return well
-      .addLiquidityQuote(Object.values(amounts))
-      .then((_quote) => {
-        quote = _quote;
-        return well.addLiquidityGasEstimate(Object.values(amounts), quote, address);
-      })
-      .then((estimate) => ({
+    try {
+      const quote = await well.addLiquidityQuote(Object.values(amounts));
+      const estimate = await well.addLiquidityGasEstimate(Object.values(amounts), quote, address);
+      return {
         quote,
         estimate
-      }));
+      };
+    } catch (error: any) {
+      Log.module("addliquidity").error("Error during quote: ", (error as Error).message);
+      return null;
+    }
   });
 
   const addLiquidityButtonClickHandler = useCallback(async () => {
     if (quote && address) {
-      const quoteAmountLessSlippage = quote.quote.subSlippage(slippage)
+      const quoteAmountLessSlippage = quote.quote.subSlippage(slippage);
       const addLiquidityTxn = await well.addLiquidity(Object.values(amounts), quoteAmountLessSlippage, address);
       const receipt = await addLiquidityTxn.wait();
       setReceipt(receipt);
@@ -152,8 +145,6 @@ export const AddLiquidity = ({ well, txnCompleteCallback, slippage, slippageSett
     checkMinAllowanceForAllTokens();
   }, [well.tokens, address, atLeastOneAmountNonzero, amounts, checkMinAllowanceForAllTokens]);
 
-  const allTokensHaveMinAllowance = useMemo(() => tokenAllowance.filter((a) => a === false).length === 0, [tokenAllowance]);
-
   const addLiquidityButtonEnabled = useMemo(
     () => address && atLeastOneAmountNonzero() && allTokensHaveMinAllowance,
     [address, atLeastOneAmountNonzero, allTokensHaveMinAllowance]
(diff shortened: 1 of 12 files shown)


projects/dex-ui/src/components/Liquidity/LiquidityRoot.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const [showSlippageSettings, setShowSlippageSettings] = useState<boolean>(false);
+  const [slippage, setSlippage] = useState<number>(0.1);
+
+  const slippageSettingsClickHandler = () => {
+    setShowSlippageSettings(true);
+  };
+
+  const slippageSettingsCloseHandler = () => {
+    setShowSlippageSettings(false);
+  };
```
0xalecks, line 26: these should be useCallbacks since they're passed as props. also maybe combine into one `toggleSlippage()` that just flips the current state. 

Or do it inline like you did for the TabButton onclick. If you do it inline, it will cause the component to re-render since it's a new function when parent re-renders

projects/dex-ui/src/components/Liquidity/LiquidityRoot.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const handleSlippageValueChange = (value: string) => {
+    console.debug(`Slippage changed: ${parseFloat(value)}`);
```
0xalecks, line 29: debug

projects/dex-ui/src/components/Liquidity/SlippagePanel.tsx
```diff
@@ -0,0 +1,82 @@
...
+type SlippagePanelProps = {
+  slippageValue: number;
+  slippageValueChanged: (value: string) => void;
+  closeButtonClicked: () => void;
+};
+
+const SlippagePanel = ({ slippageValueChanged, closeButtonClicked, slippageValue }: SlippagePanelProps) => {
+  const buttonClickHandler = () => {
+    closeButtonClicked();
+  };
```
0xalecks, line 17: no need for this, you can pass the prop to onclick directly:
`Back`

projects/dex-ui/src/components/Liquidity/SlippagePanel.tsx
```diff
@@ -0,0 +1,82 @@
...
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+import React from "react";
+import styled from "styled-components";
+
+import arrowLeft from "/src/assets/images/arrow.svg";
+
+type SlippagePanelProps = {
+  slippageValue: number;
+  slippageValueChanged: (value: string) => void;
+  closeButtonClicked: () => void;
```
0xalecks, line 11: nit on naming, this method will close the modal. IMO it should be named something like `closeModal`. using "button" or "clicked" in the name makes the assumption that it will be triggered by a button, and onClick, which neither has to be true per se

projects/dex-ui/src/components/Liquidity/SlippagePanel.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <SlippageImageSpacer>
+          <img src={arrowLeft} alt="Back" onClick={buttonClickHandler} />
+        </SlippageImageSpacer>
+        <SlippageHeaderText>Adjust Slippage</SlippageHeaderText>
+      </SlippageHeader>
+      <SlippageBody>
+        <SlippageTopSection>
+          {/* // TODO: Can we show Slippage Tolerance as a "placeholder" (per design) */}
+          {/* // TODO: Also, we need to show a trailing % */}
+          <input type="text" defaultValue={slippageValue} onChange={(e) => slippageValueChanged(e.target.value)} />
```
0xalecks, line 31: surprised this input doesn't need any styling.. haven't ran this locally. Maybe turn into into an `Input` that is styled.input`` in preparation for future design

projects/dex-ui/src/components/Liquidity/allowance.ts
```diff
@@ -0,0 +1,37 @@
+import { ERC20Token } from "@beanstalk/sdk-core";
+import { BigNumber } from "ethers";
+
+const hasMinimumAllowance = async (walletAddress: string, spender: string, token: ERC20Token, mininumAllowance: BigNumber) => {
+  const existingAllowance = await token.getContract().allowance(walletAddress, spender);
+  
+  if (!existingAllowance) {
+    throw new Error("Unable to retrieve allowance");
```
0xalecks, line 8: this error message is vague. how would we ever get into a null/undefined state? is it more likely that the call would fail and throw?

Also, forgot our previous convo, why use the contract method instead of the token's?

projects/dex-ui/src/components/Liquidity/allowance.ts
```diff
@@ -0,0 +1,37 @@
...
+      await approveTXN.wait();
+    } catch (error: any) {
+      // TODO: open notification?
+      console.log(error);
+      console.error(`Error while approving allowance: ${error.message}`);
+    }
+
+    if (allowanceCallback) {
+      allowanceCallback();
+    }
```
0xalecks, line 33: Is this pattern necessary? can't someone who is using await just do whatever after? This triggers a vague memory of a past life where we discussed this before but I don't recall :)

projects/dex-ui/src/pages/Well.tsx
```diff
@@ -35,8 +36,8 @@ export const Well = () => {
       <strong>Reserves: </strong>
       {well.reserves?.map((r, i) => `${r.toHuman("0,0.00a")} ${well.tokens?.[i].symbol}`).join(" - ")}
       <br />
-      <div>{isLoadingAllBalances ? <div>Spinner</div> : <AddLiquidity well={well} txnCompleteCallback={liquidityTxnCallback} />}</div>
-      <div>{isLoadingAllBalances ? <div>Spinner</div> : <RemoveLiquidity well={well} txnCompleteCallback={liquidityTxnCallback} />}</div>
+      <br />
+      <div>{isLoadingAllBalances ? <div>SPINNER IMAGE</div> : <LiquidityRoot well={well} txnCompleteCallback={liquidityTxnCallback} />}</div>
```
0xalecks, line 40: we have a loader component, <Spinner size={32} />

0xalecks (changes requested):

projects/dex-ui/src/components/Liquidity/AddLiquidity.tsx
```diff
@@ -7,19 +7,75 @@ import { ContractReceipt } from "ethers";
...
+  slippage: number;
+  slippageSettingsClickHandler: () => void;
 };
 
-export const AddLiquidity = ({ well, txnCompleteCallback }: AddLiquidityProps) => {
+export const AddLiquidity = ({ well, txnCompleteCallback, slippage, slippageSettingsClickHandler }: AddLiquidityProps) => {
   const { address } = useAccount();
+  const sdk = useSdk();
+  // TODO: This should be added to the main app provider I think, e.g. TokenProvider
+  const ethPrice = useEthPrice();
```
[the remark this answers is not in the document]
marshall-bean: So should I leave this for now then?

projects/dex-ui/src/components/Liquidity/AddLiquidity.tsx
```diff
@@ -32,24 +88,47 @@ export const AddLiquidity = ({ well, txnCompleteCallback }: AddLiquidityProps) =
...
+    data: quote
+    // isLoading: loadingQuote,
+    // isError: quoteError
   } = useQuery(["wells", address, amounts], async () => {
     if (!atLeastOneAmountNonzero()) {
       return null;
     }
-    return well.addLiquidityQuote(Object.values(amounts));
+
+    // TODO: Scope
```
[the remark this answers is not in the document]
marshall-bean: My mistake, tbh I am not sure what I was thinking. I'll remove it.

projects/dex-ui/src/components/Liquidity/SlippagePanel.tsx
marshall-bean: I'll go ahead and add the `styled.input` as you suggested.

projects/dex-ui/src/components/Liquidity/LiquidityRoot.tsx
marshall-bean: I can rebase and use the logger once your other PR is merged.

projects/dex-ui/src/components/Liquidity/allowance.ts
marshall-bean: First, I will remove this check. We can't get into that. It was a leftover from an earlier iteration.

Second, in this case, I can use token.getAllowance. In the other method, I have to use the contract method to specify the `gasLimit`.

projects/dex-ui/src/components/Liquidity/allowance.ts
marshall-bean: I was having issues coordinating the state, particularly once the allowance was updated. This was my solution to it. Happy to chat through it and see what you think.

projects/dex-ui/src/components/Liquidity/AddLiquidity.tsx
```diff
@@ -32,24 +88,47 @@ export const AddLiquidity = ({ well, txnCompleteCallback }: AddLiquidityProps) =
...
+    return well
+      .addLiquidityQuote(Object.values(amounts))
+      .then((_quote) => {
+        quote = _quote;
+        return well.addLiquidityGasEstimate(Object.values(amounts), quote, address);
+      })
+      .then((estimate) => ({
+        quote,
+        estimate
+      }));
```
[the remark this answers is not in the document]
marshall-bean: I'll switch to async await

projects/dex-ui/src/pages/Well.tsx
marshall-bean: ✅

projects/dex-ui/src/components/Liquidity/allowance.ts
```diff
@@ -1,35 +1,19 @@
...
 const hasMinimumAllowance = async (walletAddress: string, spender: string, token: ERC20Token, mininumAllowance: BigNumber) => {
-  const existingAllowance = await token.getContract().allowance(walletAddress, spender);
-  
-  if (!existingAllowance) {
-    throw new Error("Unable to retrieve allowance");
-  }
-  
-  return existingAllowance.gte(mininumAllowance);
+  const existingAllowance = await token.getAllowance(walletAddress, spender);
+  return existingAllowance.toBigNumber().gte(mininumAllowance);
```
0xalecks, line 7: Not sure if you need to go toBigNumber() here. I mean, you do because mininumAllowance is a BigNumber, but the only reason it's that is because you converted a TokenValue to bigNumber when you passed it in. TLDR, If you're working with TokenValues, you should never need to convert to bignumber unless your sending that value to a lower level block chain thing, like ethers or something

0xalecks (changes requested):

PR #443, opened 2023-05-07
Field: Enabled Multi Plot Transfers

uncoolzero:
Users are now able to transfer more than one plot in a single transaction.

Functionality works but some UI work still needed.


uncoolzero:
Ready for review

0xalecks (approved):

PR #450, opened 2023-05-13
Governance page updates

Bean-Sama:
- [x] Refactor the updater to fetch the delegates / delegators for the farmer. 
- [x] Fix the current account showing up as delegator for BeaNFTDao
- [x] Fix the voting power showing up on the proposals page to show the actual voting power of the farmer
- [x] Miscellaneous housecleaning


projects/ui/vite.config.ts
```diff
@@ -32,6 +32,7 @@ const CSP = buildCSP({
     '*.infura.io', // Infura RPC
     '*.bean.money', // Beanstalk APIs
     '*.snapshot.org', // Snapshot GraphQL API
+    'https://api.thegraph.com/subgraphs/name/snapshot-labs/snapshot', // Snapshot Labs GraphQL API
```
silochad, line 35: I noticed that delegators are loaded from the `snapshot-labs` API, while proposals and other Beanstalk gov info is loaded from `snapshot`.

1. Can you briefly explain why this can't all be done from the existing `snapshot` API? presumably they don't offer the delegate info?
2. This URL appears to be a hosted endpoint made available by The Graph, which is [sunsetting the hosted service](https://thegraph.com/docs/en/deploying/hosted-service/) this year. We should investigate a decentralized or self-hosted version perhaps on graph.node.bean.money.
3. Similar FYI, hub.snapshot.org will eventually require an API key. I've applied for one.

silochad (approved):

PR #451, opened 2023-05-13
Morning Auction Updates / Fixes

Bean-Sama:


silochad (approved):

projects/ui/src/state/beanstalk/sun/reducer.ts
```diff
@@ -127,27 +118,37 @@ export const selectCurrentSeason = createSelector(
...
+    // console.log('currentSeconds: ', currentSeconds.toString());
+
+    const secondsDiff = new BigNumber(maxSeconds).minus(currentSeconds);
+    // console.log('secondsDiff: ', secondsDiff.toString());
+
     // interval range [1 to 25]
     const blockDiff = currentBlock.minus(sunriseBlock);
     const interval = blockDiff.plus(1);
-    const isMorning = blockDiff.lt(25) && sunriseBlock.gt(0);
     const isMorningInterval = getIsMorningInterval(interval);
```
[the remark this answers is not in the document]
Bean-Sama: This has been deleted!

projects/ui/src/state/beanstalk/field/morning.ts
```diff
@@ -0,0 +1,145 @@
...
+    return false;
+  })();
+
+  useEffect(() => {
+    if (!shouldUpdateField) return;
+
+    const interval = setInterval(() => {
+      console.debug('[beanstalk/field/morning]: Refetching field');
+      fetchBeanstalkField();
+    }, 2000);
```
silochad, line 116: Let's move the `2000` to a constant at the top of the file, something like `FIELD_REFRESH_RATE_MS`.

projects/ui/src/state/beanstalk/field/morning.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  // If it is morning, then we fetch the soil every 4 seconds
+  useEffect(() => {
+    if (!morning.isMorning) return;
+
+    const soilUpdateInterval = setInterval(() => {
+      const now = getNowRounded();
+
+      const remaining = getDiffNow(next, now).as('seconds');
+      if (remaining % 4 === 0) {
```
silochad, line 78: Let's move the `4` to a constant at the top of the file and convert to a millisecond value, something like `MORNING_SOIL_REFRESH_RATE_MS`.

projects/ui/src/state/beanstalk/field/morning.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!first || currentSeason.lte(0)) return;
+
+    /// If we are behind 1 season, refetch
+    if (currentSeason.minus(first.season).eq(1)) {
+      console.debug('refetching temperature data...');
+      temperatureQuery.refetch({
+        season_lte: currentSeason.toNumber(),
+      });
+    }
+  }, [currentSeason, first, temperatureQuery]);
```
silochad, line 136: I'd appreciate some help understanding why we're using subgraph data here. My understanding:

`currentSeason (season.current)` = from chain
`first` = the most recently updated season in the subgraph

Later we check if `(currentSeason - first.season) = 1`, i.e. that the current season has been updated on chain but we haven't polled for new season data from the subgraph.

Followups:
- What happens if things get out of sync by more than one season? Is this possible?
- Say that `currentSeason` changes right after the gm() call, but the subgraph hasn't yet created a new season entity. Would this cause the following error? 

currentSeason updated -> season delta is now 1 -> call temperatureQuery.refetch() -> query returns the same data since subgraph hasn't yet indexed the new season -> [now what?]

If `temperatureQuery` or `first` are modified, the effect will re-run and cause a loop until the subgraph catches up.
If they aren't modified, then it seems that the state would get stuck with a difference of 1 between temperatureQuery first and currentSeason

projects/ui/src/state/beanstalk/field/morning.ts
```diff
@@ -0,0 +1,145 @@
...
+    return () => {
+      clearInterval(soilUpdateInterval);
+    };
+  }, [fetchSoil, morning.isMorning, next]);
+
+  /**
+   * Refetch the field every 2 seconds for updates if:
+   *
+   * It is not morning:
+   * - If the max temperature is not equal to the scaled temperature
```
silochad, line 92: In what case would this state be reached?

projects/ui/src/state/beanstalk/sun/morning.ts
```diff
@@ -1,116 +1,122 @@
...
+  const morningTime = useAppSelector((s) => s._beanstalk.sun.morningTime);
+  const season = useAppSelector((s) => s._beanstalk.sun.season);
+  const morning = useAppSelector((s) => s._beanstalk.sun.morning);
 
-export default function MorningUpdater() {
-  const morningTime = useSelector(selectMorningBlockTime);
-  const blockMap = useSelector(selectMorningBlockMap);
-  const morning = useSelector(selectMorning);
+  const [_, { calculate: calculateTemperature }] = useTemperature();
+  const [_soilData, { calculate: calculateNextSoil }] = useSoil();
```
silochad, line 82: Curious why the temperature is calculated for the current block, but soil is calculated for the next block during the morning

projects/ui/src/state/beanstalk/sun/morning.ts
```diff
@@ -1,116 +1,122 @@
...
+        console.debug('[beanstalk/sun/useUpdateMorning]: new block: ', {
+          temp: scaledTemp.toNumber(),
+          soil: nextSoil?.toNumber() || 'N/A',
+          blockNumber: morningResult.morning.blockNumber.toNumber(),
+          index: morningResult.morning.index.toNumber(),
+          isMorning: morningResult.morning.isMorning,
+        });
+
+        dispatch(updateScaledTemperature(scaledTemp));
+        nextSoil && dispatch(updateTotalSoil(nextSoil));
```
silochad, line 117: Should soil be split into a "total soil" and "scaled soil" like temperature is split?

silochad:
Looking good @Bean-Sama! Questions above for your consideration.

Followup questions:

1. Can you briefly explain how you tested the above peg cases (a command used to buy back above peg & any other state transitions you made would be very helpful)
2. What should users expect to see in the UI during the above peg case?

Bean-Sama:
@silochad Currently we are only showing when we are above peg to show the soil decrease amount. 
image

projects/ui/src/state/beanstalk/sun/morning.ts
Bean-Sama: so we calculate the temperature for the next block via 
`const scaledTemp = calculateTemperature(morningBlock.plus(1))`

I chose to calculate for the next morningBlock b/c the calculation requires knowledge of the temperature of the current block & the temperature of the next block. calculating 'soil' instead of 'nextSoil' would cause issues If deltaBlocks = 0, calculating soil would require, whereas if deltaBlocks = 25, we can always assume nextTemperature is the maxTemperature.

projects/ui/src/state/beanstalk/field/morning.ts
```diff
@@ -0,0 +1,145 @@
...
+   *
+   * OR
+   *
+   * If it is the morning:
+   *    We are in the 1st interval of the morning & the scaled temperature is not equal to 1.
+   *    The temperature of the 1st interval of the morning is always 1%.
```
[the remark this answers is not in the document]
Bean-Sama: Interval is defined as deltaBlocks + 1, so if currentBlock - sunriseBlock = 0, I refer to it as the 1st interval of the morning. I updated the comment, though to provide additional clarity!

projects/ui/src/state/beanstalk/field/morning.ts
Bean-Sama: Originally it was when we were transitioning out of the morning state, but this is not actually necessary anymore. Nice catch!

projects/ui/src/state/beanstalk/field/morning.ts
Bean-Sama: Yeah, this is a good point. I updated it so that we fetch the maxTemperature only when transitioning out of the morning state. This will now be called only once, to prevent these above mentioned issues.

PR #453, opened 2023-05-14
Dex UI wiggle trace

0xalecks:
adds tracing of page elements to wiggle line


projects/dex-ui/src/components/Frame/Grid/useWiggle.tsx
```diff
@@ -0,0 +1,327 @@
...
+  isPointOnBox,
+  pointsEq,
+  snapToGrid,
+  getTraceables,
+  converToSvgPoint,
+  dist,
+  getIntersection
+} from "./utils";
+
+const DEBUG = false;
```
marshall-bean, line 23: Can we use the logger here? And elsewhere in the file where we output debug information.

projects/dex-ui/src/components/Swap/TokenInput.tsx
```diff
@@ -110,7 +110,7 @@ export const TokenInput: FC<TokenInput> = ({
     }
   }, []);
 
-  if (loading) return <LoadingContainer width={width} focused={focused} />;
+  if (loading) return <LoadingContainer width={width} focused={focused} data-trace="true"ƒß/>;
```
marshall-bean, line 113: `ƒß` ?

marshall-bean (approved):
Approving, with some minor feedback that is not blocking.

projects/dex-ui/src/components/Frame/Grid/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+export const getCorners = (point: DOMPoint, obj: any) => {
+  if (!point) {
+    debugger;
+  }
+  if (point.y == obj.top) {
+    if (point.x == obj.left) {
+      // we are in the TL corner
+      return [obj.tr, obj.bl];
```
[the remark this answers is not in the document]
0xalecks: but how would you pass the `obj` variable to that? 🤔

projects/dex-ui/src/components/Swap/TokenInput.tsx
0xalecks: nice catch

projects/dex-ui/src/components/Frame/Grid/utils.ts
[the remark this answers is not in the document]
marshall-bean: For example 
```
const TOP_LEFT_CORNER = (obj) => ([obj.tr, obj.bl];
```

PR #455, opened 2023-05-15
Fix Sow With Claimable Beans

Bean-Sama:
Fixes the issue in the Sow Module where if the user is sowing using only Claimed Beans & the input token is not BEAN


silochad (approved):

silochad:
The build seems to have failed due to a failed yarn.lock update on master after a recent protocol change, I re-ran `yarn` and committed the resulting yarn.lock on this branch.

PR #459, opened 2023-05-22
Initial test setup and test cases for wells SDK

marshall-bean:
This PR contains:
- Well deployment from the SDK (instead of deploying using the wells deploy scripts)
- Initial well data and swap tests (we will add more in a future PR)
- Reused some of the test utils etc from sdk (with some changes)
- Initial GH action so tests run on every wells SDK PR
- A couple new Yarn script/commands for wells sdk tests

⚠️ I am having one weird issue around approvals. It's not blocking but as you can see it keeps me from eliminating some duplicated code.


.github/workflows/ci.sdk-wells.yaml
```diff
@@ -0,0 +1,47 @@
...
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+      - name: Launch Anvil
+        run: anvil --fork-url $ANVIL_FORK_URL --chain-id 1337 &
+        env:
+          ANVIL_FORK_URL: ${{ secrets.ANVIL_FORK_URL }}
+      - name: Build All
+        run: yarn build
+      - run: yarn sdk-wells:test
```
0xalecks, line 46:
```suggestion
      - name: Run Tests
      - run: yarn sdk-wells:test
```

package.json
```diff
@@ -24,10 +24,13 @@
     "test": "yarn sdk:test",
     "format": "yarn sdk:prettier",
     "protocol:generate": "yarn workspace @beanstalk/protocol generate",
+    "jest:clearcache": "jest --clearCache",
     "sdk:generate": "yarn workspace @beanstalk/sdk generate",
     "sdk:dev": "yarn workspace @beanstalk/sdk dev",
     "sdk:build": "yarn workspace @beanstalk/sdk build",
     "sdk:test": "jest --selectProjects sdk --silent --runInBand --",
+    "sdk-wells:test": "jest --selectProjects sdk-wells --runInBand --silent=false --",
```
0xalecks, line 32: I'd like to move away from having root level scripts like this. Can we move these to the project package.json and when we need to invoke, we can do it with `yarn workspace sdk-well test` ?

Also, if we make use of the `:` thing, which makes commands globally accessible anywhere in the workspace, you could have this in the project/packagejson/scripts:
- test: 'whatever'
- sdk-well:test: 'yarn test'

Then from anywhere, you could do `yarn sdk-well:test` and it will work.

this is just an fyi, not really sure we need to do this piece

projects/sdk-wells/test/TestUtils/BlockchainUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+  async setBEANBalance(account: string, balance: TokenValue) {
+    this.setBalance(this.sdk.tokens.BEAN, account, balance);
+  }
+
+  private getBalanceConfig(tokenAddress: string) {
+    const slotConfig = new Map();
+    slotConfig.set(this.sdk.tokens.USDC.address, [9, false]);
+    slotConfig.set(this.sdk.tokens.WETH.address, [3, false]);
+    slotConfig.set(this.sdk.tokens.BEAN.address, [0, false]);
+    return slotConfig.get(tokenAddress);
```
0xalecks, line 65: can you add DAI and USDT too please?

projects/sdk-wells/test/TestUtils/BlockchainUtils.ts
0xalecks, line 1: should we just move the one from sdk to sdk-core and share it from there?

projects/sdk-wells/test/swap.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+jest.setTimeout(30000);
+
+beforeAll(async () => {
+  await utils.resetFork();
+});
+
+const setupWell = async (wellTokens: ERC20Token[], account: string) => {
+  // Deploy test well
+  const deployment = await deployTestWellInstance(wellTokens);
```
0xalecks, line 28: very nice 😎

projects/sdk-wells/test/swap.test.ts
```diff
@@ -0,0 +1,198 @@
...
+  // await wellTokens[1].approve(testWell.address, TokenValue.MAX_UINT256);
+
+  // wellTokens.forEach(async (token) => {
+  //   await token.approve(testWell.address, TokenValue.MAX_UINT256);
+  // });
+
+  // Original code
+  await Promise.all([
+    await wellsSdk.tokens.BEAN.approve(testWell.address, TokenValue.MAX_UINT256.toBigNumber()),
+    await wellsSdk.tokens.WETH.approve(testWell.address, TokenValue.MAX_UINT256.toBigNumber())
```
0xalecks, line 55: don't need to await here

projects/sdk-wells/test/swap.test.ts
```diff
@@ -0,0 +1,198 @@
...
+  // Set initial balances for all well tokens
+  await Promise.all(
+    wellTokens.map(async (token) => {
+      await utils.setBalance(token, account, token.amount(30000));
+    })
+  );
+
+  await utils.mine();
+
+  // TODO: Why does this fail? While the original code below works?
```
0xalecks, line 40: not sure, need to try it, but i can't run it locally w/o wiping my Anvil. We can screenshare. One thing I notice is that only the for loop has .tobigNumber() at the end

0xalecks (changes requested):

projects/sdk-wells/test/swap.test.ts
marshall-bean: Are you sure? It returns a promise 🤔 And this is the same way it's done in the sdk swap tests.

package.json
marshall-bean: I want to talk more about this. I realized right now the package name is `@beanstalk/wells` and maybe this doesn't make the most sense.

projects/sdk-wells/test/TestUtils/BlockchainUtils.ts
marshall-bean: We could but I was thinking we could track that as a future task, just because it isn't in the critical path to shipping everything.

projects/sdk-wells/test/swap.test.ts
marshall-bean: Yes we can go over this in screenshare. It's very weird.

package.json
marshall-bean: Also the jest config is in the root directory :-/ so that could cause issues

projects/examples/src/wells/swap.ts
```diff
@@ -14,35 +14,60 @@ async function main() {
...
+  const wethLiquidityAmount = WETH.amount(10000);
+  const usdcLiquidityAmount = USDC.amount(10000);
 
-  const amountA = A.amount(1);
-  const amountB = B.amount(3000);
 
   // get Well object
   const well = await sdk.getWell(WELL_ADDRESS);
 
+  console.log((await well.getTokens()).map((t) => t.symbol));
```
0xalecks, line 33: console

0xalecks (approved):

PR #461, opened 2023-05-22
Dex: update UX styles

0xalecks:


marshall-bean (approved):
LGTM one small thing but not blocking

PR #469, opened 2023-05-25
Swap Router Fixes & Improvements

silochad:
# Scope

1. Adds two helper functions to create bi-directional edges for the following cases:
  a. Exchanging between two tokens in a pool
  b. Moving from an underlying token -> LP token via AddLiquidity, and vice-versa via RemoveLiquidityOneToken
2. Adds tests for the helpers to ensure both edges (A -> B _and_ B -> A) are accounted for
3. Introduces tests for some route paths where we want to ensure the router takes the most direct path. Notably:
  a. 3CRV -> underlying should always remove liquidity, and vice-versa
  b. Stables in 3pool should swap directly on 3pool, no intermediate hops
4. Add helper `expectInstanceOf` which is the same as expect.instanceOf` but with a Typescript assertion so any following context is aware of the assertion
5. Change all action constructor parameters to `public readonly`

# Rationale

**Helper functions**
1. The `graph.ts` file was becoming quite verbose, with lots of repetitive code to define edges
2. Some swap routes were missing their counterparts, for ex. we had 3CRV -> USDC but not USDC -> 3CRV mapped.

**`public readonly`**
1. Allows for tests to build an action and verify that the values stored in the instance match what you would expect based on router. example:
```
      const route = sdk.swap.router.getRoute(sdk.tokens.CRV3.symbol, symbol);
      const step = route.getStep(0).build(account, FarmFromMode.EXTERNAL, FarmToMode.INTERNAL);

      expect(route.length).toEqual(1);
      expectInstanceOf(step, RemoveLiquidityOneToken);
      expect(step._pool).toEqual(sdk.contracts.curve.pools.pool3.address); // <-- this
```
2. Javascript doesn't actually understand the concept of `private`, and `public readonly` will do effectively the same thing with respect to Typescript's warnings

**Addition of the 3CRV <> Stable route**
Some combinations of this route previously performed inefficient swaps, typically something like 3CRV -> BEAN -> USDT for example.

# Testing
Automated: see `graph.test.ts`

Manual: I rebuilt the SDK locally, set up the following swap paths, and confirmed via console logs that the correct path was selected:
- [x] DAI -> USDC
- [x] DAI -> USDT
- [x] USDC -> USDT
- [x] 3CRV -> DAI
- [x] 3CRV -> USDC
- [x] 3CRV -> USDT
- [x] reverse of each of the above

# Followup work
- Most actions use underscore prefixing for constructor params, but not all
- The AddLiquidity function accepts an array to decide which token index you're depositing, it should really be called AddLiquidityOneToken and use the same approach as RemoveLiquidityOneToken

### DIFF
--- a/projects/sdk/src/lib/farm/actions/AddLiquidity.ts
+++ b/projects/sdk/src/lib/farm/actions/AddLiquidity.ts
@@ -29,6 +29,7 @@ export class AddLiquidity extends StepClass<BasicPreparedResult> {
     });
 
     /// [0, 0, 1] => [0, 0, amountIn]
+    /// FIXME: this uses a binary approach instead of a multiplier.
     const amountInStep = this._amounts.map((k) => (k === 1 ? _amountInStep : ethers.BigNumber.from(0)));
 
     /// Get amount out based on the selected pool
--- a/projects/sdk/src/lib/farm/actions/AddLiquidity.ts
+++ b/projects/sdk/src/lib/farm/actions/AddLiquidity.ts
@@ -8,11 +8,11 @@ export class AddLiquidity extends StepClass<BasicPreparedResult> {
   public name: string = "addLiquidity";
 
   constructor(
-    private _pool: string,
-    private _registry: string,
-    private _amounts: readonly [number, number] | readonly [number, number, number],
-    private _fromMode: FarmFromMode = FarmFromMode.INTERNAL_TOLERANT,
-    private _toMode: FarmToMode = FarmToMode.INTERNAL
+    public readonly _pool: string,
+    public readonly _registry: string,
+    public readonly _amounts: readonly [number, number] | readonly [number, number, number],
+    public readonly _fromMode: FarmFromMode = FarmFromMode.INTERNAL_TOLERANT,
+    public readonly _toMode: FarmToMode = FarmToMode.INTERNAL
   ) {
     super();
   }
--- a/projects/sdk/src/lib/farm/actions/ClaimWithdrawal.ts
+++ b/projects/sdk/src/lib/farm/actions/ClaimWithdrawal.ts
@@ -5,7 +5,7 @@ import { FarmToMode } from "../types";
 export class ClaimWithdrawal extends StepClass<BasicPreparedResult> {
   public name: string = "claimWithdrawal";
 
-  constructor(private _tokenIn: string, private _season: ethers.BigNumberish, private _to: FarmToMode) {
+  constructor(public readonly _tokenIn: string, public readonly _season: ethers.BigNumberish, public readonly _to: FarmToMode) {
     super();
   }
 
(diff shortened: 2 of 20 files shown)


silochad:
FYI: The above Netlify comment seems to indicate that the UI won't get rebuilt if there's a change in the SDK code., i.e. this won't result in a deploy preview. We might want to change that behavior to also include changes to the SDK itself.

silochad:
I ran the test suite locally, only seeing this error:

```
Summary of all failing tests
 FAIL  projects/sdk/src/lib/swap/Swap.estimates.test.ts (5.031 s)
  ● Estimate › Estimate BEAN->x › estimateReverse(ETH, 3CRV)

    call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="get_dy(uint256,uint256,uint256)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.7.0)
    ```

uncoolzero (approved):
LGTM

0xalecks (approved):

silochad:
From Discord thread:

> TL;DR the changes included in #469 break the reverse estimation of a route involving [DAI/USDC/USDT]->3CRV, because the AddLiquidity and RemoveLiquidityOneToken workflow actions don't account for the workflow "direction".
> 
> For context, reverse estimation is used on the Swap page to allow users to type a desired output value into the second input and receive an estimate for how much they'd need to input.
> 
> Currently, this "estimation" is performed by assuming that there exists zero hysteresis in swaps. Ex. if I swap 1 USDC -> 0.99 DAI, if I go from DAI back to USDC I should end up with 1 USDC when everything is done. In practice this is not true due to fees and slippage, but for the sake of simplicity and speed we chose this approach for the UI. There is no deterministic way AFAIK to correctly perform the reverse estimation, the best you can get besides this zero-hysteresis assumption is an iterative approach which is highly inefficient.
> 
> In other words, when we want to estimate how much input is needed for a particular output, we start by running the entire estimation process backwards to estimate the input amount. When we finalize the transaction we run the transaction forwards to generate the final calldata.
> 
> Shipping PR #469 will cause a reversion when going to the Swap page and selecting [DAI/USDC/USDT] as input and 3CRV as output, then typing in the 3CRV box.
> 
> Given that this PR fixes a number of other inefficient routes, and the above route is still executable (you just have to set it up by only typing into the first box instead of the second), I would suggest acknowledging this bug will exist in prod and moving fwd with the PR anyway.
> 
> If a user tries to perform the route, they'll see an error message toast that reports "call revert exception" when trying to type into that output box.
> 
> As followup work, I can extend my comments on this PR to a list of future changes that would fix this issue & better structure those two actions (they need some refactors). 

#475 solves the above with an error. Merging!

PR #470, opened 2023-05-25
Transfer: Farm balance warning

silochad:
**DO NOT MERGE** until approval of #469.

# Scope
Adds a red warning to the Transfer page if the user is delivering assets to the Farm balance of some other account.

# Rationale
The recipient will need access to the Farm balance to use the assets. Transferring to an account like a Coinbase exchange wallet will cause a loss of funds.


hellofromguy (approved):

uncoolzero (approved):
looks good to me!

PR #518, opened 2023-06-28
Fix contract functions, pt. 2

silochad:
UI & SDK builds now passing (though underlying naming & some logic needs reworking according to Silo V3 spec).

- [x] Refactor EventProcessor to support new events, update EventManager
- [x] Switch Silo & Field updaters in the UI to fetching & processing events via the SDK's event processor, remove the UI one
- [x] First-pass simplification of the EventProcessor, followup PRs will reduce things (especially types) down further, for ex. there are few types for Deposit crates
- [x] Get rid of duplicate UI code related to event processing
- [x] Add helper for migrating from BNJS <> ethers BN <> TokenValue


silochad:
Requesting retroactive review

PR #519, opened 2023-06-28
Event processor: Field

silochad:
Reintroduce the Field event processor; migrate from the previous BigNumberJS & object approach to ethers.BigNumber and Maps.

Reintroduce helper function parsePlots to split plots into harvestable & non-harvestable. Add test for this.

Use the processor in the farmer/field/updater and wrangle the ethers.BigNumbers into BigNumberJS which is what the UI components expect currently. Not pretty but functional. Tested manually on some of my accounts.

`yarn run jest processor.test`

### DIFF
--- a/projects/sdk/src/lib/events/processor.ts
+++ b/projects/sdk/src/lib/events/processor.ts
@@ -1,9 +1,9 @@
 import { ethers } from "ethers";
 import { Token } from "src/classes/Token";
 import {
-  // SowEvent,
-  // HarvestEvent,
-  // PlotTransferEvent,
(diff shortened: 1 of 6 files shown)


projects/sdk/src/lib/events/processor.ts
```diff
@@ -83,6 +76,8 @@ export class EventProcessor {
 
     // Field
     this.plots = initialState?.plots || new Map();
+
+    return this;
```
silochad, line 80: reverting this change in a followup pr

PR #521, opened 2023-06-28
Refactor silo balances, add stems support to SDK

silochad:
- [x] Reduce the TokenSiloBalance entity to one layer with `amount`, `bdv`, `deposits` (formerly crates)
- [x] Rename crates -> deposits in most important locations in `silo.ts` and `utils.ts`
- [x] Refactor season -> stem and add support for calculating grown stalk from stems

Future work: rename crates to deposits in other locations (Withdraw.ts, etc.)


silochad:
Requesting retroactive review

PR #524, opened 2023-06-30
Crate refactors

silochad:
The UI has a lot of duplicate code that's also in the SDK. I migrated the farmer/silo updater to use the SDK (and thus stems), but that's created a lot of downstream issues. This PR fixes those & attempts to migrate some of the main forms to the SDK. I'm removing a lot of legacy code which was ported to the SDK & TokenValue that still exists in the UI w/ BigNumbers.


silochad:
Requesting retroactive review

PR #528, opened 2023-07-03
Tweaks, stalk struct, other cleanup and misc fixes

silochad:
- Tweaks the onboarding flow to use mowAndMigrateNoDeposits, adds animation
- Switches the UI to use a new stalk struct which contains total, base, and grown. Remove calculations from in UI components since this is already done in the updater
- Fix a few display bugs
- Make tables work with both stems & seasons-based deposit data
- Rework revitalized stalk calculations
- Begin work on earned bean vesting


silochad:
Requesting retroactive review

PR #544, opened 2023-07-08
Rewards bar refactor

silochad:
Silo V3 introduces mowing for individual tokens, which breaks the previous paradigm in the Claim Rewards dialog. The dialog also really sucks. I went ahead and fixed both of these at the same time, and updated the logic to use `mowMultiple` when necessary.

image

# Future work
- Add tooltips to explain why certain things are disabled
- Form isn't resetting correctly after submit
- Roundoff errors with local enroot calculation, maybe use static call to calculate instead?


silochad:
Requesting retroactive review

PR #604, opened 2023-08-14
add owner() function

Brean0:
Adds owner() function. Although there is no `onlyOwner()` modifier, some nft marketplaces uses owner() to give permissions to edit the collection details. Currently set to the BCM as a constant, due to the fact that changing the owner would require a BIP, and thus would be redundant to add additional functionality to transfer ownership of the beanstalk deposits.


Brean0:
not applicable, owner() already exists within beanstalk (and is pointed to BCM).

PR #606, opened 2023-08-15
Basin UI - Upgrades

uncoolzero:


projects/sdk-wells/src/lib/liquidity/AddLiquidityETH.ts
```diff
@@ -0,0 +1,86 @@
...
+        [wrapEth, wethTransfer, syncWell],
+        ethAmount.toBlockchain()
+    ]);
+
+    return { steps: [...tokenTransfers, pipe], ethAmount: ethAmount };
+  }
+
+  async doGasEstimate(well: Well, amounts: TokenValue[], quote: TokenValue, account?: string, overrides: TxOverrides = {}): Promise<TokenValue> {
+    const { steps, ethAmount } = this.generateSteps(well, amounts, quote, account || this.ZERO_ADDRESS);
+    const overrideOptions = { ...overrides, value: ethAmount.toBigNumber(), gasLimit: 5000000 };
```
0xalecks, line 75: are you sure we need to include a gasLimit here? This value is used in the `.estimateGas()` method which doesn't actually send a transaction or use gas, if my understanding is correct

0xalecks (changes requested):

0xalecks (approved):

PR #735, opened 2023-12-23
Add wallet connect

Bean-Sama:
Add Wallet connect to list of available wallets


projects/dex-ui/src/utils/wagmi/Client.tsx
```diff
@@ -81,13 +81,13 @@ export const client: any = createClient({
...
-    //   chains,
-    //   options: {
-    //     projectId: "TODO",
-    //     showQrModal: true
-    //   }
-    // }),
+    new WalletConnectConnector({
+      chains,
+      options: {
+        projectId: import.meta.env.VITE_WALLET_CONNECT_PROJECT_ID,
```
0xalecks, line 87: We should check that this env file exists and either throw an error (breaking the whole app), or conditionally not load the connector if the id isn't set.

0xalecks (changes requested):

0xalecks (approved):

PR #736, opened 2023-12-29
BIP-X: BDV Decrease

nickkatsios:
## Summary

Implemented a new Convert type that allows any user to decrease a Deposit’s BDV if the Recorded BDV (the BDV stored on-chain with the Deposit) is greater than the Current BDV (the number of tokens in the Deposit * the current BDV of that token).

## Problem:

Currently, the Recorded BDV of a given Deposit is asynchronous with its Current BDV. For example, if a user Deposits BEANETH into the Silo and the USD price of ETH falls significantly (and thus the BDV of BEANETH decreases), the Recorded BDV does not decrease, i.e., the user “locked in” the greater BDV at the time of Deposit. In practice, this may mean Beanstalk is over crediting BDV to Deposits relative to their value to the system.

Beanstalk only supports Convert types that do not decrease the BDV of the Deposit.

## Solution

- Implemented a new Convert type `ANTI_LAMBDA_LAMBDA` , callable by anyone, that Converts on behalf of a user.
- Added an additional parameter ```account``` in ```LibConvert.convert(…)``` , corresponding to the account to decrease the deposit of.
-  Created a new decoder ```antiLambdaConvert(bytes memory self)``` function that additionally decodes the account parameter.
- Replaced ```msg.sender``` in ```ConvertFacet``` ```.convert()``` ,```. _withdrawTokens()``` , ```. _depositTokensForConvert()``` functions and all needed places with ```account```.
- Challenged the new functionallity with tests in ```test/Convert.test.js```
- Added mock bdv selectors in ```MockSiloFacet``` to artificially manipulate the bdv of a deposit in tests.

### DIFF
--- a/protocol/contracts/beanstalk/silo/ConvertFacet.sol
+++ b/protocol/contracts/beanstalk/silo/ConvertFacet.sol
@@ -15,10 +15,9 @@ import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
 import {SafeCast} from "@openzeppelin/contracts/utils/SafeCast.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {LibConvert} from "contracts/libraries/Convert/LibConvert.sol";
-import "hardhat/console.sol";
 
 /**
- * @author Publius, Brean, DeadManWalking
+ * @author Publius, Brean, deadmanwalking
  * @title ConvertFacet handles converting Deposited assets within the Silo.
  **/
 contract ConvertFacet is ReentrancyGuard {
@@ -83,15 +82,6 @@ contract ConvertFacet is ReentrancyGuard {
         * and decreaseBDV is true */
         (toToken, fromToken, toAmount, fromAmount, account, decreaseBDV) = LibConvert.convert(convertData);
 
-        console.log("Inside ConvertFacet");
-        console.log("Convert decoded");
-        console.log("toToken: %s", toToken);
-        console.log("fromToken: %s", fromToken);
-        console.log("toAmount: %s", toAmount);
-        console.log("fromAmount: %s", fromAmount);
-        console.log("decreaseBDV: %s", decreaseBDV);
-        console.log("account convert is called on: %s", account);
-        
         require(fromAmount > 0, "Convert: From amount is 0.");
 
         // replace account with msg.sender if no account is specified
@@ -101,8 +91,6 @@ contract ConvertFacet is ReentrancyGuard {
         LibSilo._mow(account, toToken);
 
         // withdraw the tokens from the deposit 
-        // calls LibTokenSilo.removeDepositFromAccount()
-        // removeDepositFromAccount subtracts the amount and bdv from the deposit
         (grownStalk, fromBdv) = _withdrawTokens(
             fromToken,
             stems,
@@ -115,17 +103,13 @@ contract ConvertFacet is ReentrancyGuard {
         uint256 newBdv = LibTokenSilo.beanDenominatedValue(toToken, toAmount);
 
         // if we have used the anti-lambda-lamda convert, 
-        // we need to decrease the bdv of the new deposit
+        // we need to update the bdv of the new deposit
         if(decreaseBDV) {
 	        toBdv = newBdv;
         } else {
         // else keep the max of the two bdvs
 	        toBdv = newBdv > fromBdv ? newBdv : fromBdv;
         }
-
-        console.log("withdrawTokensCompleted. Results:");
-        console.log("fromBdv: %s", fromBdv);
-        console.log("toBdv Final: %s", toBdv);
 
         toStem = _depositTokensForConvert(toToken, toAmount, toBdv, grownStalk, account);
 
@@ -143,7 +127,6 @@ contract ConvertFacet is ReentrancyGuard {
             stems.length == amounts.length,
             "Convert: stems, amounts are diff lengths."
         );
-        console.log("Inside ConvertFacet _withdrawTokens");
         LibSilo.AssetsRemoved memory a;
         uint256 depositBDV;
         uint256 i = 0;
@@ -225,8 +208,7 @@ contract ConvertFacet is ReentrancyGuard {
             "Convert: Not enough tokens removed."
         );
         LibTokenSilo.decrementTotalDeposited(token, a.tokensRemoved, a.bdvRemoved);
-        console.log("withdrawTokens: bdv removed %s", a.bdvRemoved);
-        console.log("withdrawTokens: tokens removed %s", a.tokensRemoved);
+
         LibSilo.burnStalk(
             account,
             a.stalkRemoved.add(a.bdvRemoved.mul(s.ss[token].stalkIssuedPerBdv))
@@ -244,12 +226,6 @@ contract ConvertFacet is ReentrancyGuard {
     ) internal returns (int96 stem) {
         require(bdv > 0 && amount > 0, "Convert: BDV or amount is 0.");
        
-        console.log("Inside ConverFacet: _depositTokensForConvert");
-        console.log("Account to deposit to: %s", account);
-        console.log("bdv to re add: %s", bdv);
-        console.log("amount of tokens to re add: %s", amount);
-
-
         //calculate stem index we need to deposit at from grownStalk and bdv
         //if we attempt to deposit at a half-season (a grown stalk index that would fall between seasons)
         //then in affect we lose that partial season's worth of stalk when we deposit
--- a/protocol/contracts/beanstalk/silo/ConvertFacet.sol
+++ b/protocol/contracts/beanstalk/silo/ConvertFacet.sol
@@ -83,6 +83,12 @@ contract ConvertFacet is ReentrancyGuard {
         * and decreaseBDV is true */
         (toToken, fromToken, toAmount, fromAmount, account, decreaseBDV) = LibConvert.convert(convertData);
 
+        console.log("Inside ConvertFacet");
+        console.log("Convert decoded");
+        console.log("toToken: %s", toToken);
+        console.log("fromToken: %s", fromToken);
+        console.log("toAmount: %s", toAmount);
+        console.log("fromAmount: %s", fromAmount);
         console.log("decreaseBDV: %s", decreaseBDV);
         console.log("account convert is called on: %s", account);
         
@@ -108,8 +114,6 @@ contract ConvertFacet is ReentrancyGuard {
         // calculate the bdv of the new deposit
         uint256 newBdv = LibTokenSilo.beanDenominatedValue(toToken, toAmount);
 
-        console.log("fromBdv: %s", fromBdv);
-
         // if we have used the anti-lambda-lamda convert, 
         // we need to decrease the bdv of the new deposit
         if(decreaseBDV) {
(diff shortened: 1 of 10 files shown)


protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -74,37 +74,54 @@ contract ConvertFacet is ReentrancyGuard {
         nonReentrant
         returns (int96 toStem, uint256 fromAmount, uint256 toAmount, uint256 fromBdv, uint256 toBdv)
     {
-        address toToken; address fromToken; uint256 grownStalk; address account;
+        address toToken; address fromToken; uint256 grownStalk; address account; bool decreaseBDV;
+
+        /** @dev account and decreaseBDV are initialized at the start of LibConvert.convert()
```
Brean0, line 79: comment is good, but imo not needed as line 87 is self explanatory 
- could move to LibConvert to inform future developers.

protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -74,37 +74,54 @@ contract ConvertFacet is ReentrancyGuard {
...
+        * and decreaseBDV is true */
+        (toToken, fromToken, toAmount, fromAmount, account, decreaseBDV) = LibConvert.convert(convertData);
 
-        (toToken, fromToken, toAmount, fromAmount) = LibConvert.convert(convertData);
-        
         require(fromAmount > 0, "Convert: From amount is 0.");
 
-        LibSilo._mow(msg.sender, fromToken);
-        LibSilo._mow(msg.sender, toToken);
+        // replace account with msg.sender if no account is specified
```
Brean0, line 87: Personally like capitalizing comments and ending in periods.

protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -74,37 +74,54 @@ contract ConvertFacet is ReentrancyGuard {
...
+        if(account == address(0)) account = msg.sender;
+
+        LibSilo._mow(account, fromToken);
```
Brean0, line 90: given that we now have two converts where to and from token are the same, think we can add logic to check whether toToken != fromToken and omit the 2nd mow.

protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -74,37 +74,54 @@ contract ConvertFacet is ReentrancyGuard {
...
-            fromAmount
+            fromAmount,
+            account
         );
 
         // calculate the bdv of the new deposit
         uint256 newBdv = LibTokenSilo.beanDenominatedValue(toToken, toAmount);
 
-        toBdv = newBdv > fromBdv ? newBdv : fromBdv;
+        // if we have used the anti-lambda-lamda convert,
```
Brean0, line 105: lets update the comment to be more generalized. Additionally, "we" should refer to the developer rather than the user using the function. 
```suggestion
               // if `decreaseBDV` flag is not enabled, set toBDV to the max of the two bdvs
```

protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -74,37 +74,54 @@ contract ConvertFacet is ReentrancyGuard {
...
+            account
         );
 
         // calculate the bdv of the new deposit
         uint256 newBdv = LibTokenSilo.beanDenominatedValue(toToken, toAmount);
 
-        toBdv = newBdv > fromBdv ? newBdv : fromBdv;
+        // if we have used the anti-lambda-lamda convert, 
+        // we need to update the bdv of the new deposit
+        if(decreaseBDV) {
```
Brean0, line 107: we can refactor this such that decreaseBDV doesn't need to be checked on every convert.  if NewBdv > fromBDV, decreaseBDV does not need to be checked (`or` statements will break early if the first conditional is 1, and `and` statements will break early if the first conditional is 0).

Although minor, its good to get into the habit of checking whether checks can be omitted for future use.

```suggestion
        toBdv = (newBdv > fromBdv || decreaseBDV)  ? newBdv : fromBdv;
```

protocol/contracts/beanstalk/silo/ConvertFacet.sol
```diff
@@ -203,10 +221,11 @@ contract ConvertFacet is ReentrancyGuard {
         address token,
         uint256 amount,
         uint256 bdv,
-        uint256 grownStalk // stalk grown previously by this deposit
+        uint256 grownStalk, // stalk grown previously by this deposit
+        address account // account to update the deposit (used in bdv decrease)
     ) internal returns (int96 stem) {
         require(bdv > 0 && amount > 0, "Convert: BDV or amount is 0.");
-
+
```
Brean0, line 228: remove trailspace

protocol/contracts/libraries/Convert/LibConvert.sol
```diff
@@ -62,6 +64,10 @@ library LibConvert {
         } else if (kind == LibConvertData.ConvertKind.UNRIPE_TO_RIPE) {
             (tokenOut, tokenIn, amountOut, amountIn) = LibChopConvert
                 .convertUnripeToRipe(convertData);
+        } else if (kind == LibConvertData.ConvertKind.ANTI_LAMBDA_LAMBDA) {
+	        (tokenOut, tokenIn, amountOut, amountIn, account) = LibLambdaConvert
+                .antiConvert(convertData);
+	        decreaseBDV = true;
```
Brean0, line 70: instead of setting decreaseBDV here, libLamdaConvert.antiConvert should return false, to match the current format 

lets add it in LibConvertData.antiLambdaConvert instead, given that we don't want to allow the user to specify whether to decreaseBDV.

protocol/contracts/libraries/Convert/LibLambdaConvert.sol
```diff
@@ -26,4 +30,26 @@ library LibLambdaConvert {
         tokenOut = tokenIn;
         amountOut = amountIn;
     }
+
+    /** 
+     * @notice This function returns the full input for use in anti-lamda convert
+     * In anti lamda convert, any user can convert on behalf of an account 
+     * to update a dedposit's bdv.
```
Brean0, line 37: deposits

protocol/contracts/libraries/Convert/LibLambdaConvert.sol
```diff
@@ -26,4 +30,26 @@ library LibLambdaConvert {
...
+        pure
+        returns (
+            address tokenOut,
+            address tokenIn,
+            uint256 amountOut,
+            uint256 amountIn,
+            address account
+        )
+    {
+        (amountIn, tokenIn , account) = convertData.antiLambdaConvert();
```
Brean0, line 51: remove space between tokenIn and comma

protocol/contracts/libraries/Silo/LibTokenSilo.sol
```diff
@@ -259,9 +259,8 @@ library LibTokenSilo {
             return removedBDV;
         }
         // Full remove
+        // deletes the deposit from storage
```
Brean0, line 262: combine 261 and 262 imo

Brean0 (commented):
review 1. great work, mostly stylistic changes

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: agree on that. fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/b40ac94815c889d4318f7bfe76039ef55e5f1069)

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/b40ac94815c889d4318f7bfe76039ef55e5f1069)

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: You are right this sounds a bit awkward. Fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/b40ac94815c889d4318f7bfe76039ef55e5f1069)

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: Saves a bit of gas and keeps it clean as well. Will definitely keep this in mind. Fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/b40ac94815c889d4318f7bfe76039ef55e5f1069)

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/b40ac94815c889d4318f7bfe76039ef55e5f1069)

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: and the rest [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/8747da303de09d3562588586af2045210af1eda8)

protocol/contracts/libraries/Convert/LibConvert.sol
nickkatsios: You mean something like [this](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/10fc00c5eba6b41bb6e26b1eb8b020f7b458478d) ?

nickkatsios: You mean something like [this](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/10fc00c5eba6b41bb6e26b1eb8b020f7b458478d) ?

protocol/contracts/beanstalk/silo/ConvertFacet.sol
nickkatsios: agree on this one moved [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/10fc00c5eba6b41bb6e26b1eb8b020f7b458478d) as a note

protocol/contracts/libraries/Convert/LibLambdaConvert.sol
nickkatsios: fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/10fc00c5eba6b41bb6e26b1eb8b020f7b458478d)

protocol/contracts/libraries/Convert/LibLambdaConvert.sol
nickkatsios: fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/10fc00c5eba6b41bb6e26b1eb8b020f7b458478d)

nickkatsios: fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/10fc00c5eba6b41bb6e26b1eb8b020f7b458478d)

protocol/contracts/libraries/Silo/LibTokenSilo.sol
nickkatsios: of course, fixed [here](https://github.com/BeanstalkFarms/Beanstalk/pull/736/commits/4a67fe7ca1d7cb2b4ac4da1922879d96ad079221)

PR #755, opened 2024-01-23
Subgraph: Pre Replant DeltaB Values

cujowolf:
Include deltaB values from pre replant in both the subgraph and analytics graph query.


soilking:
Closing because these changes are included in #851

PR #761, opened 2024-02-01
Bean UI - Update Governance page

uncoolzero:
Expanded our governance page to support older proposals, EBIPs and more

Updated the list of votes table with ENS support


projects/ui/src/functions/ebipdata/ebipdata.ts
```diff
@@ -0,0 +1,71 @@
...
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(output),
+        };
+      };
+    };
+
+    return {
+      statusCode: 403
```
0xalecks, line 58: Should seems like the wrong response code. 403 is for unauthorized. Did you mean 404 not found, or 400 bad request?

projects/ui/src/functions/oldbipdata/oldBipList.ts
0xalecks, line 1: do we need to keep these oldXXXX functions?

0xalecks (changes requested):

projects/ui/src/functions/ebipdata/ebipdata.ts
uncoolzero: updated to 404

projects/ui/src/functions/oldbipdata/oldBipList.ts
uncoolzero: good call, removed the votes as we aren't really using them

projects/ui/src/functions/oldbipdata/oldbipdata.ts
```diff
@@ -0,0 +1,71 @@
...
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(output),
+        };
+      };
+    };
+
+    return {
+      statusCode: 404
```
0xalecks: this should be a 500 if it's an error

0xalecks (approved):

PR #791, opened 2024-02-29
Tractor sdk

funderbrker:
# Tractor SDK
This PR provides basic SDK functionality for interacting with Tractor. It also expands current functionality of pipeline.

## Features
- Easy composition of sequences of actions
- Common reward actions
- Standardized handling of byte-level data
- Tooling to assist operators in simulating Blueprints

### What this is not
- A standardized set of Blueprints that will be offered.


funderbrker:
## Status

- Core SDK design in place. Compiles and runs. Test reverts contract side.
- https://github.com/BeanstalkFarms/Pipeline/issues/9 issue replicated here. Seeing an extra 32 bytes in clipboard data constructed through SDK. Consensus is that these bytes are expected and whitepaper should be updated to represent them.

## Todo
- Refine SDK design. Some structure can be slightly shuffled to improve usability.
- Implement standard Blueprint and sub-blueprints
- Implement full suite of tests that cover both Tractor SDK and Contracts. Deprecate existing js Tractor tests.
- UI design & implementation (out of scope for this PR)

PR #814, opened 2024-04-02
Subgraph: Modified basin daily snapshot to trigger at 9 am PT/12pm EST

soilking:
Updates the Basin daily snapshot time (`wellDailySnapshots`) to occur at 9 am PT/12pm EST. This is primarily used by the bots in the Basin Discord. This change should go out at the same time as the bot modification (below).

There was no clear way to mock block timestamps, so I tested by deploying to subgraph studio and observing at what block numbers the new snapshots were taken. https://thegraph.com/studio/subgraph/beandev/playground

Related to [Beanstalk-Py-15](https://github.com/BeanstalkFarms/Beanstalk-Py/pull/15)


projects/subgraph-basin/src/utils/Well.ts
```diff
@@ -197,7 +197,9 @@ export function incrementWellWithdraw(wellAddress: Address): void {
 
 export function checkForSnapshot(wellAddress: Address, timestamp: BigInt, blockNumber: BigInt): void {
   // We check for the prior period snapshot and then take one if needed
-  let dayID = dayFromTimestamp(timestamp) - 1;
+  // Schedule the "day" to begin at 9am PT/12pm ET.
+  // Future work could include properly adjusting this when DST occurs.
+  let dayID = dayFromTimestamp(timestamp, 8 * 60 * 60) - 1;
```
cujowolf, line 202: Shouldn't this be `-7 * 60 * 60` based on the update to `dayFromTimestamp`?

cujowolf (changes requested):

projects/subgraph-basin/src/utils/Well.ts
soilking: Setting it as `-7 * 60 * 60` would make the day begin 7 hours later (`dayFromTimestamp` now thinks 0:00 is actually 17:00), which would occur at 12am PT. In this case we want the day to begin at 9am which is either 16 hours later (-16) or 8 hours earlier (+8). +8 was chosen as to prevent an overly delayed update on whatever day this is deployed

soilking:
This was deployed, and should be merged

BeanstalkFarmsOperations (approved):

PR #817, opened 2024-04-05
Subgraph: Peg Crosses due to non-bean token price fluctuation

soilking:
Adds `BlockHandler` which runs on each block, and checks whether a peg cross has occurred due to eth (or 3crv) price fluctuation. This begins after the deployment of `BeanstalkPrice` contract on block `17978222`.


soilking:
Closing because these changes are included in #851

PR #823, opened 2024-04-10
remove fundraiser facet and its references

funderbrker:
this concept and code is no longer in use. presumably too stale to be used anyway.


Brean0:
lgtm!

Brean0 (approved):

PR #824, opened 2024-04-11
Excessive price fix. BEAN/USD rather than USD/BEAN

Brean0:


protocol/contracts/libraries/Well/LibWell.sol
```diff
@@ -219,14 +219,22 @@ library LibWell {
...
+    function getBeanTknPriceFromTwaReserves(address well) internal view returns (uint256 price) {
         AppStorage storage s = LibAppStorage.diamondStorage();
         // s.twaReserve[well] should be set prior to this function being called.
-        // 'price' is in terms of reserve0:reserve1.
         if (s.twaReserves[well].reserve0 == 0 || s.twaReserves[well].reserve1 == 0) {
             price = 0;
         } else {
-            price = s.twaReserves[well].reserve0.mul(1e18).div(s.twaReserves[well].reserve1);
+            // fetch the bean index from the well in order to properly return the bean price.
+            if(getBeanIndexFromWell(well) == 0) {
```
BrendanSanderson, line 233: Add a space after the if

protocol/contracts/libraries/Well/LibWell.sol
```diff
@@ -219,14 +219,22 @@ library LibWell {
         s.twaReserves[well].reserve1 = 1;
     }
 
-    function getWellPriceFromTwaReserves(address well) internal view returns (uint256 price) {
+    /**
+     * @notice returns the price in terms of TKN/BEAN. 
+     * (if eth is 1000 beans, this function will return 1000e6);
+     */
+    function getBeanTknPriceFromTwaReserves(address well) internal view returns (uint256 price) {
```
BrendanSanderson, line 226: Can you spell out “Token” instead of “Tkn”

protocol/contracts/libraries/Well/LibWell.sol
Brean0: fixed: https://github.com/BeanstalkFarms/Beanstalk/commit/d2a52383fcdf8985fb1ce59aeee0b04cbb5dadc1#diff-78cd81288a9e2da2d97a68af6ca1deb23ebe60d5e57b49edb44c690e98470964R233

protocol/contracts/libraries/Well/LibWell.sol
Brean0: Fixed: d2a52383fcdf8985fb1ce59aeee0b04cbb5dadc1

PR #826, opened 2024-04-13
Foundry testing

Brean0:
supports more tests.


pizzaman1337:
@Brean0 github auto-deleted the foundry-testing branch after I merged, we likely need to keep this unless you have this work somewhere else?

PR #832, opened 2024-04-16
rm depot

funderbrker:


protocol/test/foundry/utils/DepotDeployer.sol
```diff
@@ -1,25 +0,0 @@
...
-pragma abicoder v2;
-
-import {Utils, console} from "test/foundry/utils/Utils.sol";
-
-/**
- * @title DepotDeployer
- * @author Brean
- * @notice Test helper contract to deploy Depot.
- */
-contract DepotDeployer is Utils {
```
Brean0, line 14: Could we add this back? Pipeline will need to be deployed in order for the depotFacet to work properly. I prefer that depot is deployed, but don't think it matters too much given beanstalk does not use the standalone depot contract.

Brean0 (changes requested):

protocol/test/foundry/utils/DepotDeployer.sol
funderbrker: I will add it back. I deleted it because it was not being used in any of the tests, but it sounds like it is a reference on how to deploy depot.

PR #846, opened 2024-04-29
rm marketplace v1

funderbrker:
removes legacy v1 code for handling listings + orders

removes `v2` naming scheme. what was previously labelled as v2 will now appear the same as v1 (no version label)


funderbrker:
new plan

PR #851, opened 2024-04-30
Bean subgraph - 2.2.0

soilking:
Contains various fixes and upgrades, these are the new features:
- Peg crosses according to eth price fluctuation in the absence of a swap or new season (for both pre-replant UniswapV2 and now our Well)
- Historical price, deltaB, twaPrice, and twaDeltaB across all pools and aggregated
- Locked Beans from unripe since Replant, and its effect on the L2SR
- Dewhitelisted pools are omitted from overall Bean price calculation

The UI should be updated to show the prices and deltas for twa vs. instantaneous, but this can be addressed in a separate PR/deployment.


soilking:
Note that this branch does currently include a change to one of the ui charts, this needs to be either completed or reverted prior to merging.

uncoolzero (approved):

BeanstalkFarmsOperations (approved):

PR #856, opened 2024-05-07
7to8

funderbrker:


funderbrker:
TODO:
review and replace `BASE_FEE_CONTRACT`. Investigate sunrise reward impact.

protocol/contracts/libraries/LibFertilizer.sol
```diff
@@ -230,4 +229,33 @@ library LibFertilizer {
...
+     * Among others, `isContract` will return false for the following
+     * types of addresses:
+     *
+     *  - an externally-owned account
+     *  - a contract in construction
+     *  - an address where a contract will be created
+     *  - an address where a contract lived, but was destroyed
+     * ====
+     */
+    function isContract(address account) internal view returns (bool) {
```
Brean0, line 251: Believe in solidity 0.8.0 and up, the evm now exposes the `code` member, which allows you to determine whether the address is a contract. https://docs.soliditylang.org/en/latest/units-and-global-variables.html?highlight=address+initial#members-of-address-types. This definitely works as well but defer to you on implmentation.

Brean0 (commented):
Great job! Apologies on the delay, wanted to be thorough in review.

protocol/contracts/libraries/LibFertilizer.sol
funderbrker: I didn't want to shake the tree on the intricacies of how this vs .code might behave differently during construction or after destruction. current open zeppelin version does not offer any form of contract checking.

protocol/contracts/beanstalk/diamond/DiamondCutFacet.sol
```diff
@@ -3,7 +3,7 @@
 */
 
 pragma experimental ABIEncoderV2;
-pragma solidity =0.7.6;
+pragma solidity ^0.8.20;
```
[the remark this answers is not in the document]
funderbrker: there are 3 diamond standards. all are still valid but have different types of gas optimizations. the diamond system used by beanstalk is a modified from any of the standards. the versioning change should not affect the diamond behavior and i suggest we do not alter our implementation unless there is something we specifically want to acheive.

PR #860, opened 2024-05-10
shipping & receiving

funderbrker:
refactor the distribution system on new bean mints so that it is more clean, has more fine grained control, allows dynamic calculation of distribution proportions, and can be updated without modifying the facets. also generalizes fields to work with the new distribution system in a way that reduces future need to modify facets.


protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,154 @@
...
+        uint256 beansToShip
+    ) private pure {
+        for (uint256 i; i < shipmentPlans.length; i++) {
+            // Do not modify amount for streams with 0 points. They either are zero or have already been set.
+            if (shipmentPlans[i].points == 0) continue;
+            shipmentAmounts[i] = (beansToShip * shipmentPlans[i].points) / totalPoints; // round down
+        }
+    }
+
+    //////////////////////////////////// SHIPPING PLAN GETTERS ////////////////////////////////////
```
Brean0, line 128: Could we put the shipping plan getters in a seperate facet? The main benefit is that the shipping plans can be updated without updating the season facets.

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,154 @@
...
+    // These getters do *not* need to live inside of the Beanstalk Diamond. However, this initial
+    // set of getters is placed here for simplicity. It is possible to add a new Shipping Route
+    // and dynamic Plan without modifying Beanstalk Contract logic.
+
+    /**
+     * @notice Get the current points and cap for Barn shipments.
```
Brean0, line 134: Could we add comments on why the cap is this value? (i.e fertilizer is capped by the amount of fert sold etc etc)

protocol/contracts/beanstalk/init/InitDistribution.sol
```diff
@@ -0,0 +1,49 @@
+/*
+ SPDX-License-Identifier: MIT
+*/
+
+pragma solidity ^0.8.20;
+
+import {AppStorage} from "contracts/beanstalk/AppStorage.sol";
+import {C} from "contracts/C.sol";
+import {LibDiamond} from "contracts/libraries/LibDiamond.sol";
```
Brean0, line 9: LibDiamond looks to be unused here, is that right?

protocol/contracts/beanstalk/init/InitDistribution.sol
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author funderbrker
+ * @notice Initializes logistics and shipment routes.
+ */
+contract InitDistribution {
+    AppStorage internal s;
+
+    function init() external {
+        s.shipmentRoutes.push(
```
Brean0, line 22: This works as an init, but ideally we should probably have functionality somewhere that allows people to add, modify, and remove shipment routes. We should also have events that correspond to these state changes. In my head, this looks pretty similar to how LibWhitelistedTokens.sol is set up. Could we implment a library that does that and use that here?

protocol/contracts/beanstalk/init/InitDistribution.sol
```diff
@@ -0,0 +1,49 @@
...
+            Storage.ShipmentRoute(
+                address(this),
+                bytes4(keccak256("siloReceive(uint256,bytes)")),
```
Brean0, line 25: This works, but personally I prefer using the `x.siloReceive.selector` just because it seems to be more foolproof. Could we implment a `IShipmentPlan` interface (that `Distrubution.sol` or if its decided to make the shipment plan functions placed in a `ShipmentPlanFacet` inherits).

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
```diff
@@ -0,0 +1,112 @@
...
+ * @author funderbrker
+ * @notice Holds the functions responsible for receiving Bean shipments after mints. These
+ *  functions must live inside of the Beanstalk Diamond. If new receiving components are needed,
+ *  this contract and its containing Facet will need to be updated.
+ * @dev An alternative design could remove the need for the generalized receive() entry function
+ *  and instead require the shipping route to define the selector of its own corresponding receive
+ *  function. However, both designs will require a Facet cut if a new receive function is needed,
+ *  so this design was chosen for additional clarity.
+ */
+contract Receiving is ReentrancyGuard {
```
Brean0, line 32: Since `Receiving` inherits `ReentrancyGuard`, `Oracle.sol` does not need to inherit `ReentrancyGuard` anymore. Could we remove that from oracle.sol?

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,154 @@
...
+
+        Storage.ShipmentRoute[] memory shipmentRoutes = s.shipmentRoutes;
+        ShipmentPlan[] memory shipmentPlans = new ShipmentPlan[](shipmentRoutes.length);
+        uint256[] memory shipmentAmounts = new uint256[](shipmentRoutes.length);
+        uint256 totalPoints;
+        (shipmentPlans, totalPoints) = getShipmentPlans(shipmentRoutes);
+
+        // May need to calculate individual stream rewards multiple times, since
+        // they are dependent on each other and each has a cap.
+        for (uint256 i; i < shipmentRoutes.length; i++) {
```
Brean0, line 47: Is this implmented without the loop in the case that a cap is hit currently?

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,154 @@
...
+            bool capExceeded;
+            // Calculate the amount of rewards to each stream. Ignores cap and plans with 0 points.
+            getBeansFromPoints(shipmentAmounts, shipmentPlans, totalPoints, beansToShip);
```
Brean0, line 50: is `shipmentAmounts` passed by reference or pass by value? From my understanding it was passed by value.

Brean0 (commented):
Small review (just general comments). Great work!

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: non-value types (like arrays) are always passed by reference

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: I am unsure what you mean.
If no cap is hit in the first pass it will break out without performing unnecessary loops

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
```diff
@@ -0,0 +1,112 @@
...
+
+import {C} from "contracts/C.sol";
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import {AppStorage, Storage} from "contracts/beanstalk/AppStorage.sol";
+import {LibFertilizer} from "contracts/libraries/LibFertilizer.sol";
+import {ReentrancyGuard} from "contracts/beanstalk/ReentrancyGuard.sol";
+
+/**
+ * @notice Constraints of how many Beans to send to a given route at the current time.
+ * @param points Weight of this shipment route relative to all routes.
```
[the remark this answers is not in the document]
funderbrker: Yes, but I still haven't decided on a proper precision..

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
funderbrker: It is inheriting reentrancy guard because the AppStorage variable `s` is defined there. if `s` is also declared here it will conflict and break. i do not understand the design decision to place the declaration in reentrancy guard. could use some insight here

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: They aren't worthy of a facet. Instead we could

- put them in a separate contract that is inherited into the season facet for cleaner code separation
or
- put them in an external contract that is outside beanstalk and immutable. then redeploy externally and update shipping route plan getter contract address on any future updates.

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
funderbrker: `Oracle.sol` also uses this class variable

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: Definitely.

Thankfully this system of updating can be much simpler than LibWhitelisted tokens.

Added here https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/08614d4814895da7c59ddd8c24b0eb489ca783f0#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833R120

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: fixed here
https://github.com/BeanstalkFarms/Beanstalk/commit/08614d4814895da7c59ddd8c24b0eb489ca783f0#diff-b0451bea09abb2293197eb00e388806cf4949fc7fd2952a7c926693780a93f60R34

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
```diff
@@ -0,0 +1,112 @@
...
+ *  this contract and its containing Facet will need to be updated.
+ * @dev An alternative design could remove the need for the generalized receive() entry function
+ *  and instead require the shipping route to define the selector of its own corresponding receive
+ *  function. However, both designs will require a Facet cut if a new receive function is needed,
+ *  so this design was chosen for additional clarity.
+ */
+contract Receiving is ReentrancyGuard {
+    using SafeCast for uint256;
+
+    function receiveShipment(Storage.Recipient recipient, uint256 amount, bytes memory data) internal {
```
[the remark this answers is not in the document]
funderbrker: added 
https://github.com/BeanstalkFarms/Beanstalk/pull/860/files#diff-6c2813749618ebaa2ded478935db44e4d09054278e07e6ed5a595a1a6f7ab178R31

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
```diff
@@ -0,0 +1,112 @@
...
+        // `s.earnedBeans` is an accounting mechanism that tracks the total number
+        // of Earned Beans that are claimable by Stalkholders. When claimed via `plant()`,
+        // it is decremented. See {Silo.sol:_plant} for more details.
+        s.earnedBeans = s.earnedBeans + amount.toUint128();
+
+        // Mint Stalk (as Earned Stalk).
+        // Stalk is created here because only Beans that are allocated to the Silo receive Stalk.
+        s.silo.stalk = s.silo.stalk + (amount * C.STALK_PER_BEAN);
+
+        s.siloBalances[C.BEAN].deposited = s.siloBalances[C.BEAN].deposited + uint128(amount);
```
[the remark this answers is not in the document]
funderbrker: yup

https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/bbc0486ed0591c1b30cb75365c393b9efa03d353#diff-6c2813749618ebaa2ded478935db44e4d09054278e07e6ed5a595a1a6f7ab178R56

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: i decided to place them in a contract that lives outside of beanstalk

https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/bbc0486ed0591c1b30cb75365c393b9efa03d353#diff-cf3f4898f41eee71e7b37cb1f5e642c532b4efd2ce1ee23de550f167e61ea6b5R25

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: added for all plan functions around here
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/bbc0486ed0591c1b30cb75365c393b9efa03d353#diff-cf3f4898f41eee71e7b37cb1f5e642c532b4efd2ce1ee23de550f167e61ea6b5R35

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: removed
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/08614d4814895da7c59ddd8c24b0eb489ca783f0#diff-b0451bea09abb2293197eb00e388806cf4949fc7fd2952a7c926693780a93f60L9

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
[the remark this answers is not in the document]
Brean0: lgtm!

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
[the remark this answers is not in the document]
Brean0: lgtm

protocol/contracts/beanstalk/init/InitDistribution.sol
Brean0: Beautiful! works for me!

protocol/contracts/beanstalk/init/InitDistribution.sol
Brean0: ❤️

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
Brean0: Got it, lets leave for now and reevaluate later.

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
Brean0: dope, ty

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
Brean0: Got it, this is great! Realized that the worst case would be `O(n^2)`, which checks out here. Would the shipment plans need to be re-calculated in the loop, given they are dependent on each other? Currently the current system redistributes the extra beans proportional to the new points %. Ideally the system would have more flexibility in controlling its distribution. 

Imagine a hypothetical setup where we have a silo, active field, barn, and legacy field, with the distribution of 45%, 45%, 5%, 5%. Respectively. Say the beanstalk wants to setup distribution such that: 
    - if the barn is paid off, extra bean seignorage goes to the legacy field. 
    - if the legacy field is paid off, extra bean seignorage goes to the barn. 
    - if the active field is paid off, extra bean seignorage goes to the silo.

With the current system, It doesn't seem possible to implement static numbers to fit this functionality. For example if the active field was paid off, the new distribution would be 80%, 10%, 10% respectively.

protocol/contracts/beanstalk/init/InitDistribution.sol
```diff
@@ -0,0 +1,55 @@
...
+        );
+
+        shipmentRoutes[2] = Storage.ShipmentRoute(
+            shipmentPlanner,
+            ShipmentPlanner.getBarnPlan.selector,
+            Storage.Recipient.Barn,
+            bytes("")
+        );
+
+        BEANSTALK.delegatecall(abi.encodeWithSelector(beanstalk.setShipmentRoutes.selector, shipmentRoutes));
```
Brean0, line 53: Can we call the function directly rather than delegate calling?

Also, while we can use the beanstalk address directly, the benefit of using `IBeanstalk(address.this)` is that any beanstalk fork can initialize their beanstalk using the existing init contract deployed rather than having to redeploy their own `InitDistribution.sol`.
```suggestion
        IBeanstalk(address(this)).setShipmentRoutes(shipmentRoutes);
```

protocol/contracts/beanstalk/init/InitDistribution.sol
```diff
@@ -0,0 +1,55 @@
...
+ * @author funderbrker
+ * @notice Initializes logistics and shipment routes.
+ */
+contract InitDistribution {
+    address constant BEANSTALK = 0xC1E088fC1323b20BCBee9bd1B9fC9546db5624C5;
+
+    AppStorage internal s;
+    IBeanstalk beanstalk = IBeanstalk(BEANSTALK);
+
+    address shipmentPlanner = address(new ShipmentPlanner());
```
Brean0, line 27: typically we try to create vanity addresses for external contracts. Could we have a hardhat script script that deploys the shipment planner, and put the address of that as an init input?

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,164 @@
...
+            }
+            totalPoints += shipmentPlans[i].points;
+        }
+    }
+
+    /**
+     * @notice Replaces the entire set of ShipmentRoutes with a new set.
+     * @dev Changes take effect immediately and will be seen at the next sunrise mint.
+     */
+    function setShipmentRoutes(Storage.ShipmentRoute[] calldata shipmentRoutes) external {
```
Brean0, line 156: Could we also have view functions for the shipment route in storage? Defer to you on whether we want to also include functions that expose the individual values as well as the struct.

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: setShipmetRoutes is an external function that is only accessible to the owner. So using delegate call keeps the msg.sender address as the owner. 

Would it be better to set it to only be external but only callable by Beanstalk address such that it can only be called through Inits and DiamondCuts?

Are there any examples of other external functions that are protected similarly?

Brean0: Got it. Totally slipped my mind re msg.sender. In cases where an external function is access locked, we've historically placed the logic in a library (where the external function is a wrapper around it + access control). Then in the init contract we call the library. This definitely works as well unchanged.

protocol/contracts/beanstalk/field/FieldFacet.sol
```diff
@@ -167,89 +169,102 @@ contract FieldFacet is Invariable, ReentrancyGuard {
...
-            delete s.podListings[index];
-            emit PodListingCancelled(LibTractor._user(), index);
-        }
+        harvestablePods = s.fields[fieldId].harvestable.sub(index);
+
+        // TODO: Verify this works. Solicit review.
+        // _cancelPodListing(LibTractor._user(), fieldId, index);
+        address(this).delegatecall(
+            abi.encodeWithSelector(IBeanstalk.cancelPodListing.selector, fieldId, index)
+        );
```
funderbrker, line 188: @Brean0 

this is a call from one facet to a different facet's public function. i am unsure about the implications of this design. i could not find any other examples in the codebase of similar behavior. wdyt about this approach?

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: That functionality is not supported by this design. Once a system is paid off it will cap itself or set its points to 0, which will result in all other output streams receiving a larger portion of the total bean mints.

It is not clear how a fully circularly dependent system could be designed, but certain features could likely be achieved. However, what your describing is out of scope atm.

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: partially achieved here, although i was unable to call the beanstalk contract object directly 

https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d431b0731dbf151a4f810b4e7bdfab1a728a323f#diff-b0451bea09abb2293197eb00e388806cf4949fc7fd2952a7c926693780a93f60R56-R57

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: added here
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/212e19e040586cc4b4b2e4404b85099613344ab3#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833R153-R159

protocol/contracts/beanstalk/field/FieldFacet.sol
Brean0: lgtm - ideally `_cancelPodListing` would've been a function from a library but given it was written prior, in this context it makes sense to me.

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: updated here:
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/1c333eadf4106d25342ca7ede416bea79453891a#diff-b0451bea09abb2293197eb00e388806cf4949fc7fd2952a7c926693780a93f60L32-L35

vanity deploy script & address can be set outside this PR

protocol/contracts/beanstalk/field/FieldFacet.sol
funderbrker: moved it into a lib here
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/0575d331733ccf2c10cdf48c0140458a3b8b922d#diff-d1df16bc4f6f3b8f37c60a5dab5717e751f8a2986236fc936965d2ecd0f0bdf6R198

i am not sure what blocked me from doing that in the first place, but agree this is the better solution.

protocol/contracts/beanstalk/init/InitDistribution.sol
funderbrker: got it
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d5c6b3cc0e9749cb496e1de9a924392da46cd8c4#diff-b0451bea09abb2293197eb00e388806cf4949fc7fd2952a7c926693780a93f60R60-R62

funderbrker:
LGF

protocol/contracts/beanstalk/init/InitDistribution.sol
Brean0: lgtm

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,172 @@
...
+     */
+    function getShipmentRoutes() public view returns (Storage.ShipmentRoute[] memory) {
+        return s.shipmentRoutes;
+    }
+
+    /**
+     * @notice Replaces the entire set of ShipmentRoutes with a new set.
+     * @dev Changes take effect immediately and will be seen at the next sunrise mint.
+     */
+    function setShipmentRoutes(Storage.ShipmentRoute[] calldata shipmentRoutes) public {
```
Brean0, line 164: can we mark this as `external` ? `setShipmentRoutes` is not used internally.

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,172 @@
...
+                shipmentPlans[i] = ShipmentPlan({points: 0, cap: 0});
+            }
+            totalPoints += shipmentPlans[i].points;
+        }
+    }
+
+    /**
+     * @notice Gets the current set of ShipmentRoutes.
+     */
+    function getShipmentRoutes() public view returns (Storage.ShipmentRoute[] memory) {
```
Brean0, line 156: Can we mark this as `external`?

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
```diff
@@ -0,0 +1,119 @@
...
+        uint256 amount,
+        bytes memory data
+    ) internal {
+        if (recipient == Storage.ShipmentRecipient.Silo) {
+            siloReceive(amount, data);
+        } else if (recipient == Storage.ShipmentRecipient.Field) {
+            fieldReceive(amount, data);
+        } else if (recipient == Storage.ShipmentRecipient.Barn) {
+            barnReceive(amount, data);
+        }
```
Brean0, line 42: can we add a comment on the bottom indiciating that new ShipmentRecipient enums should be added here?

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * @notice Receive Beans at the Field. The next `amount` Pods become harvestable.
+     * @dev Amount should never exceed the number of Pods that are not yet Harvestable.
+     * @param amount Amount of Beans to receive.
+     * @param data Encoded uint256 containing the index of the Field to receive the Beans.
+     */
+    function fieldReceive(uint256 amount, bytes memory data) private {
+        uint256 fieldId = abi.decode(data, (uint256));
```
Brean0, line 72: could we add a check here that `fieldId` doesn't exceed `s.fieldCount`?

protocol/contracts/ecosystem/ShipmentPlanner.sol
```diff
@@ -0,0 +1,69 @@
...
+
+interface IBeanstalk {
+    function isFertilizing() external view returns (bool);
+    function totalUnfertilizedBeans() external view returns (uint256);
+
+    function isHarvesting(uint256 fieldId) external view returns (bool);
+    function totalUnharvestable(uint256 fieldId) external view returns (uint256);
+}
+
+uint256  constant BARN_POINTS = 333_333_333_333_333_333;
```
Brean0, line 23: 1) remove extra space
2) should the constants lie inside `ShipmentPlanner`?

protocol/contracts/beanstalk/field/FieldFacet.sol
```diff
@@ -167,89 +183,122 @@ contract FieldFacet is Invariable, ReentrancyGuard {
...
      * @notice Returns the number of Pods that are not yet Harvestable. Also known as the Pod Line.
+     * @param fieldId The index of the Field to query.
      */
-    function totalUnharvestable() public view returns (uint256) {
-        return s.field.pods.sub(s.field.harvestable);
+    function totalUnharvestable(uint256 fieldId) public view returns (uint256) {
+        return s.fields[fieldId].pods - s.fields[fieldId].harvestable;
+    }
+
+    function isHarvesting(uint256 fieldId) public view returns (bool) {
```
Brean0, line 291: could we add natspec here?

protocol/contracts/ecosystem/ShipmentPlanner.sol
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @notice Get the current points and cap for Field shipments.
+     * @dev The Field cap is the amount of outstanding Pods unharvestable pods.
+     * @param data Encoded uint256 containing the index of the Field to receive the Beans.
+     */
+    function getFieldPlan(
+        bytes memory data
+    ) external view returns (ShipmentPlan memory shipmentPlan) {
+        uint256 fieldId = abi.decode(data, (uint256));
```
Brean0, line 56: can we check that fieldId < s.fieldCount here as well?

protocol/contracts/ecosystem/ShipmentPlanner.sol
```diff
@@ -0,0 +1,69 @@
...
+
+uint256  constant BARN_POINTS = 333_333_333_333_333_333;
+uint256  constant FIELD_POINTS = 333_333_333_333_333_333;
+uint256  constant SILO_POINTS = 333_333_333_333_333_333;
+
+/**
+ * @title ShipmentPlanner
+ * @notice Contains getters for retrieving ShipmentPlans for various Beanstalk components.
+ * @dev Lives as a standalone immutable contract. Updating shipment plans requires deploying
+ *      a new instance and updating the ShipmentRoute planContract addresses help in AppStorage.
```
Brean0, line 31: 1. remove tab
2. could we add that new plan must be view/pure functions?

protocol/contracts/ecosystem/ShipmentPlanner.sol
```diff
@@ -0,0 +1,69 @@
...
+ */
+contract ShipmentPlanner {
+    address constant BEANSTALK = 0xC1E088fC1323b20BCBee9bd1B9fC9546db5624C5;
```
Brean0, line 34: can we make this have a constructor with an immutable variable BEANSTALK? This would allow beanstalk forks to be spun up without changing `shipmentPlanners` (would be nice for an potential L2 migration)

protocol/contracts/beanstalk/sun/SeasonFacet/Weather.sol
```diff
@@ -205,9 +205,11 @@ contract Weather is Sun {
         uint256 newHarvestable;
 
         // Pay off remaining Pods if any exist.
-        if (s.field.harvestable < s.rain.pods) {
-            newHarvestable = s.rain.pods - s.field.harvestable;
-            s.field.harvestable = s.field.harvestable.add(newHarvestable);
+        if (s.fields[s.activeField].harvestable < s.rain.pods) {
```
Brean0, line 208: CC @pizzaman1337  on how this effects gen-flood

Brean0 (commented):
Review II. Just stylistic changes. Good stuff

protocol/contracts/beanstalk/field/FieldFacet.sol
```diff
@@ -167,89 +183,129 @@ contract FieldFacet is Invariable, ReentrancyGuard {
...
+    /**
+     * @notice Set the active Field. Only the active field is accrues Soil.
+     * @param fieldId ID of the Field to set as active. ID is the Field Number.
+     */
+    function setActiveField(uint256 fieldId, uint32 temperature) public fundsSafu noSupplyChange noNetFlow {
+        LibDiamond.enforceIsOwnerOrContract();
+        require(fieldId < s.fieldCount, "Field: Field does not exist");
+        s.activeField = fieldId;
+
+        // Reset weather.
```
funderbrker, line 235: @Brean0 the weather and case system is a too much for me to dive into rn. does this make logic make sense for resetting the weather when updating fields?

protocol/contracts/beanstalk/AppStorage.sol
```diff
@@ -520,4 +547,11 @@ struct AppStorage {
     mapping(IERC20 => uint256) internalTokenBalanceTotal;
     uint256 fertilizedPaidIndex;
     uint256 plenty;
+    // Distribution.
+    Storage.ShipmentRoute[] shipmentRoutes;
+    // Fields.
+    uint256 activeField;
```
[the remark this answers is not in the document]
funderbrker: being restructured in #861 already

protocol/contracts/beanstalk/AppStorage.sol
```diff
@@ -421,16 +445,18 @@ contract Storage {
  * @param paused True if Beanstalk is Paused.
  * @param pausedAt The timestamp at which Beanstalk was last paused.
  * @param season Storage.Season
- * @param field Storage.Field
+ * @param fields mapping of Field ID to Storage.Field.
+ * @param activeField ID of the active Field.
```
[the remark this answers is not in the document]
funderbrker: all scrambled in #861 anyway. will fix there

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,172 @@
...
+     * @dev Determines which routes are active and how many Beans they will receive.
+     * @dev GetPlan functions should never fail/revert. Else they will have no Beans allocated.
+     */
+    function getShipmentPlans(
+        Storage.ShipmentRoute[] memory shipmentRoutes
+    ) private view returns (ShipmentPlan[] memory shipmentPlans, uint256 totalPoints) {
+        shipmentPlans = new ShipmentPlan[](shipmentRoutes.length);
+        for (uint256 i; i < shipmentRoutes.length; i++) {
+            (bool success, bytes memory returnData) = shipmentRoutes[i].planContract.staticcall(
+                abi.encodeWithSelector(shipmentRoutes[i].planSelector, shipmentRoutes[i].data)
```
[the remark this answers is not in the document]
funderbrker: good callout, but in this particular case it is incompatible bc we need to hold the selector in the plan object. `abi.encodeCall` requires a "function pointer" instead of bytes4. i do not see a way to convert bytes4 into function pointer.

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
funderbrker: i believe this is checked automatically in 0.8 and an appropriate panic code will be thrown. an explicit check could let us set a msg and improve debugging, but i think it would be pretty clear here.

funderbrker: this is checked automatically in 0.8, right? and an appropriate panic code thrown. or do you feel it could benefit from a more useful revert message?

protocol/contracts/beanstalk/field/FieldFacet.sol
funderbrker: @pizzaman1337 may also have insight?

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,172 @@
...
+                }
+            }
+            // If no cap exceeded, amounts are final.
+            if (!capExceeded) break;
+        }
+
+        // Ship it.
+        for (uint256 i; i < shipmentAmounts.length; i++) {
+            if (shipmentAmounts[i] == 0) continue;
+            // Success is not verified here. Receive functions should be designed to never revert.
```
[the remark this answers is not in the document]
funderbrker: yes
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833L79

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,172 @@
...
+        uint256 beansToShip
+    ) private pure {
+        for (uint256 i; i < shipmentPlans.length; i++) {
+            // Do not modify amount for streams with 0 points. They either are zero or have already been set.
+            if (shipmentPlans[i].points == 0) continue;
+            shipmentAmounts[i] = (beansToShip * shipmentPlans[i].points) / totalPoints; // round down
+        }
+    }
+
+    function hashShipmentRoute(
```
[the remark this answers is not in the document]
funderbrker: left over from a previous design iteration. removed
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833L117

protocol/contracts/beanstalk/field/FieldFacet.sol
funderbrker: added
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-d1df16bc4f6f3b8f37c60a5dab5717e751f8a2986236fc936965d2ecd0f0bdf6R301-R304

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
funderbrker: https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-6c2813749618ebaa2ded478935db44e4d09054278e07e6ed5a595a1a6f7ab178R27

protocol/contracts/beanstalk/sun/SeasonFacet/Receiving.sol
funderbrker: added
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-6c2813749618ebaa2ded478935db44e4d09054278e07e6ed5a595a1a6f7ab178R75

protocol/contracts/ecosystem/ShipmentPlanner.sol
funderbrker: added
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-cf3f4898f41eee71e7b37cb1f5e642c532b4efd2ce1ee23de550f167e61ea6b5R61

protocol/contracts/ecosystem/ShipmentPlanner.sol
funderbrker: updated here. do not need to store the address anywhere
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-cf3f4898f41eee71e7b37cb1f5e642c532b4efd2ce1ee23de550f167e61ea6b5R38-R40

protocol/contracts/ecosystem/ShipmentPlanner.sol
funderbrker: yes
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-cf3f4898f41eee71e7b37cb1f5e642c532b4efd2ce1ee23de550f167e61ea6b5R28-R29

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: here
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833R141

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
funderbrker: fixed
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833R133

protocol/contracts/ecosystem/ShipmentPlanner.sol
funderbrker: fixed spacing and moved inside contract

https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/d65b6bacc7e6d96105a48c759af2f5222b0ccb26#diff-cf3f4898f41eee71e7b37cb1f5e642c532b4efd2ce1ee23de550f167e61ea6b5R32-R34

protocol/contracts/beanstalk/sun/SeasonFacet/Distribution.sol
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * @title Distribution
+ * @author funderbrker
+ * @notice Handles shipping of new Bean mints.
+ */
+contract Distribution is Receiving {
+    using SafeCast for uint256;
+
+    /**
+     * @notice Emitted during Sunrise when Beans are distributed to the Field, the Silo, and Fertilizer.
```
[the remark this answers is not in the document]
funderbrker: fixed 
https://github.com/BeanstalkFarms/Beanstalk/pull/860/commits/85fecec784e9e817f0eb2d48f719d88c0aa19ba3#diff-a3ff8fea8dfdc60875a144d7d0f1fb3b1e5f0b82f2cbcb1331f2fca38d73f833R22

PR #876, opened 2024-05-21
Update readme to include known working node/npm versions

pizzaman1337:
Adds note in `protocol/README.md` about compatible node/npm versions.

The other spacing modifications appear to be from auto-format.


funderbrker (approved):
ty

PR #888, opened 2024-05-24
Rm old silos

funderbrker:


funderbrker:
@Brean0 seeds also needs to be removed here. but is also deeply engrained in the tests
https://github.com/BeanstalkFarms/Beanstalk/pull/888/files#diff-84e80fbf84a48c6d151b08d628223fe89d3ef4cb8fcd9e6b92c3b1da76252f8aL119

PR #952, opened 2024-07-09
Pod Marketplace Subgraph + UI Fixes

soilking:
This should be deployed/merged along with `subgraph-beanstalk2.3.0`. Another PR will be necessary in the Beanstalk-Py project to support the updated pod order event.

Bug Fixes:
- Fix for orders/listings that did not appear, both on the chart and in individual users' market
- Fix showing incorrect place in line on market history tab
- Fix missing events due to exceeding query limit
- Fix `fillOrder` transactions not fully filling an order

Other changes:
- Pod listing partial fill, then cancelled, is now `CANCELLED_PARTIAL` instead of `CANCELLED`
- Refactor market orders to be in terms of beans ordered rather than pods
    - `PodOrder.podAmount` removed
- Add tracking for amounts of expired beans/pods
- Tracking amount of beans in active pod orders
- Refactor listing remainingAmount to also indicate the amount at expiry/cancellation
- Add `Plot.beansPerPod` field. This indicates the price at which a plot was acquired whether via sowing or on the marketplace
- Add `Plot.updatedAtBlock` field. This will allow for easier computation of a dollar cost basis
- Added `logIndex` into `historyId` field


projects/ui/src/components/Market/PodsV2/Modules/MarketGraph.tsx
```diff
@@ -581,7 +582,7 @@ const Graph: FC<GraphProps> = ({
         hideTooltip();
       }
     } else if (params.listingID) {
-      const index = listings.findIndex((l) => l.id === params.listingID);
+      const index = listings.findIndex((l) => l.id.slice(43) === params.listingID);
```
Space-Bean, line 585: Could you provide a comment here as to what 43 means?

Space-Bean (approved):

PR #978, opened 2024-07-26
Multi-manifest setup for subgraph deployments

soilking:
Restructures the usage of subgraph.yaml files such that multiple can exist within the same subgraph project. This enables deploying the same subgraph mapping with a different set of activated handlers.

Renames fields in the beanstalk subgraph:
`window` -> `emaWindow`
`from` -> `fromFarmer`
`to` -> `toFarmer`


Space-Bean (approved):
lgtm!

PR #994, opened 2024-08-01
adds support for deploying upgradable well implementations and s2

Brean0:


nickkatsios:
lgtm, merging to integrate with the rest of the workflow

PR #1014, opened 2024-08-08
Verify can't withdraw without losing grown stalk through pipeline convert

pizzaman1337:
Fixes issue reported here: https://codehawks.cyfrin.io/c/2024-05-beanstalk-the-finale/s/425
Uses test provided (with some slight cleanup)


Brean0 (approved):
lgtm

PR #1016, opened 2024-08-09
M-01 Add support for oracle timeout

pizzaman1337:
Original issue: https://codehawks.cyfrin.io/c/2024-05-beanstalk-the-finale/s/476

For stablecoins, after 4 hour, the oracle facet would return a price of 0 due to the timeout time being non-configurable. This adds the ability to configure the timeout for oracle Implementations. For volatile assets, a 4 hour timeout is generally sufficient, and for stablecoins, a 4 day timeout.

Note this PR branched off the branch that removes the chainlink registry feature, as this builds on top of that.


protocol/contracts/beanstalk/storage/System.sol
```diff
@@ -389,9 +389,10 @@ struct Migration {
  * gaugePoint and liquidityWeight implementations.
  */
 struct Implementation {
-    address target;
+    address target; // 20 bytes
     bytes4 selector;
     bytes1 encodeType;
+    uint32 timeout; // 4 bytes
```
Brean0, line 395: we should make this a generic bytes so that future external implementations can have generic byte data to be added to its oracle (i.e a generalized oracle that can be used for multiple tokens.

PR #1018, opened 2024-08-11
Update temp increase if ladder in LibDibbler

nickkatsios:
- Update the morning temperature increase to: 
- ### $log_{3.5}\left(0.1 \cdot \text{block} + 1\right)$
- By updating the values in the if ladder located in `LibDibbler.sol`


Brean0:
lgtm, great work.

PR #1041, opened 2024-08-19
subgraph-beanstalk misc fixes

soilking:
- Workaround for bug in `TotalGerminatingBalanceChanged` event emission
- Fix double counting of some stalk in vapy calculation
- Fix improper precision in chopped bdv calculation


Space-Bean (approved):
lgtm!

Issue #1036, opened 2024-08-19
RFC: Fork Migration System


funderbrker:
## Authors

funderbrker, Ben Weintraub, Brendan Sanderson, Beanstalk Farms

## Summary

Implement a Fork Migration System that allows Farmers to migrate assets from one or more Beanstalks (Source Beanstalks) to another (a Destination Beanstalk) under conditions defined by the Destination Beanstalk.

## Context

Forks and upgrades both allow protocols to be modified:
* Upgrades: Occurs when there is some native mechanism in the software that allows the rules of the existing instantiation of the system to be changed. As there is still only 1 instance of the system, participants are forced into the upgrade.
* Forks: Occurs when a set of participants decide to change the rules of the protocol such that a new instance of the system is created. Each participant has the option to keep using the old system or switch to the new one.

Forks give participants complete agency while upgrades minimize fragmentation.
 
A powerful property of governance in networks like Bitcoin and Ethereum is that there is a guarantee that the rules of the system will not change without participants opting into the changes. Bitcoin miners who did not update their software to Bitcoin Cash remained on Bitcoin. Ethereum nodes that did not update their software to facilitate the fork that reorganized Ethereum post-DAO hack remained on ETH Classic. When juxtaposed against smart contract upgrades, where the rules can be changed for participants who did not opt into it, social consensus seems preferable for a base money.

Furthermore, because smart contracts do not have a strong relationship with network validators, it is harder for smart contracts to coordinate a reorganization via social consensus in the instance of a flaw (e.g., Ethereum did not reorganize post-Beanstalk exploit). Therefore, even though there is friction in the form of lost network effects, it seems the optimal governance model for smart contracts is to make them non-upgradable and to make migrating to future implementations (or forks) as frictionless as possible such that there can be a Lindy effect around safe implementations to use without slowing down innovation.

While this solution likely creates more overhead for developers to support backward compatibility, it also provides more similar properties for users of smart contracts as it does for users of systems based on social consensus.

### Centralized Governance

Since the April 2022 governance exploit, Beanstalk governance has been off-chain. In addition to being fundamentally misaligned with the ethos of Beanstalk, off-chain governance has numerous censorship risks.

### Counting Votes

It is unclear how to count votes in the fairest way possible. Because of the ability to use infinite addresses, without a unique identity system there is no way to ensure that each participant only votes once. 

[...]

PR #1042, opened 2024-08-20
Fork System

funderbrker:
# Fork System
The Fork System enables an opt-in system of upgrading Beanstalk. Anyone, including Beanstalk Farms, can implement descendant forks that allow users to opt into the upgrade by migrating existing parent assets into the child.

## RFC
https://github.com/BeanstalkFarms/Beanstalk/issues/1036

## Design of Destination
A source fork has no knowledge of child forks, however each child fork must specifically whitelist and implement logic for the parent forks it will support. This implementation demonstrates one way that a child fork could be designed to accept migrated assets, however, it is not the only way it can be done. Each child fork is responsible for deciding how to mint assets for migrating users. The possible design space is infinite and is not fully captured by this PR, however destination logic modifications should fit inside of the standardized interfaces demonstrated in this PR.
 
## Implementation notes
Although a decent amount of tech debt had to be addressed throughout the codebase while implementing this system, the core of the changes can be found in
- `TransmitOutFacet.sol`
- `LibTransmitOut.sol`
- `TransmitInFacet.sol`
- `LibTransmitIn.sol`
- `TransmitToFork.t.sol`


protocol/contracts/beanstalk/field/FieldFacet.sol
```diff
@@ -264,42 +271,44 @@ contract FieldFacet is Invariable, ReentrancyGuard {
...
     }
 
     /**
-     * @notice Returns the number of Pods that have ever been Harvested.
+     * @notice Returns the number of Pods that have ever been Harvested or Burned.
      * @param fieldId The index of the Field to query.
      */
-    function totalHarvested(uint256 fieldId) public view returns (uint256) {
-        return s.sys.fields[fieldId].harvested;
+    function totalProcessed(uint256 fieldId) public view returns (uint256) {
```
pizzaman1337, line 285: Is there a way to get total that were harvested?

protocol/contracts/beanstalk/field/FieldFacet.sol
```diff
@@ -191,10 +198,14 @@ contract FieldFacet is Invariable, ReentrancyGuard {
         // are harvestable.
         harvestablePods = s.sys.fields[fieldId].harvestable.sub(index);
```
pizzaman1337, line 199: Is there a test that verifies if you harvest a slashed/burned plot, then anyone can harvest it and also this function returns 0? I am hopefully misunderstanding, but it seems like currently this function might not return 0 for slashed plots and beans might get sent out for them

protocol/test/foundry/sun/Sun.t.sol
```diff
@@ -105,36 +109,36 @@ contract SunTest is TestHelper {
...
+        // // no beans should be minted.
+        // if (deltaB <= 0) {
+        //     assertEq(
+        //         initialBeanBalance - C.bean().balanceOf(address(bs)),
+        //         0,
+        //         "invalid bean minted -deltaB"
+        //     );
+        //     assertEq(bs.totalSoil(), soilIssued, "invalid soil @ -deltaB");
+        //     assertEq(bs.totalUnharvestable(0), podsInField, "invalid pods @ -deltaB");
+        // }
```
pizzaman1337, line 141: Why was this test commented out?

protocol/contracts/beanstalk/field/FieldFacet.sol
funderbrker: harvesting a slashed/burned plot does not return 0 here, but should not send out beans. i feel this is tested in TransmitToFork.t.sol, but not sure that it specifically checks the bean balance to confirm nothing is being sent.

PR #1061, opened 2024-09-03
Fix: Mint bean sided liquidity to the L2 BCM

nickkatsios:
- Upon migration, the L2 BCM will add liquidity to the BEAN:WETH, BEAN:wstETH and BEAN:USDC Wells in the ratios exported from the L1 state.
- The non bean tokens ie WETH, wstETH and USDC are all obtained from removing liquidity from the existing L1 wells and transferring them through the L2 native bridge. 
- Considering that the BEAN token will be a new contract and L1 beans cannot be migrated, only reissued, the BCM will also need L2 Beans to be able to add the bean sided liquidity in the L2 wells. 
- This pr adds the following: 
- Passes the bean liquidity amounts to `ReseedBean` and mints the beans to the BCM in `mintBeansToBCM(bean, wethBeans, wstEthBeans, stableBeans);`


Brean0 (commented):
lgtm!

PR #1063, opened 2024-09-04
Fix LibWhitelist.verifyOracleImplementation() is inconsistent with LibUsdOracle.sol

pizzaman1337:
T1MOH593 report [L-04] fix.

LibUsdOracle uses address(this) if target address is not specified:
However in sanity check during whitelist it will make call to address(0) which always succeeds instead of real sanity check (`function verifyOracleImplementation`)
This performs a similar check in `LibWhitelist.verifyOracleImplementation`.

Would've liked to have written a test for it, but it seems there's currently no way to fully exercise this code path due to no matching public selector available for use.

### DIFF
--- a/protocol/contracts/libraries/Oracle/LibUsdOracle.sol
+++ b/protocol/contracts/libraries/Oracle/LibUsdOracle.sol
@@ -141,11 +141,10 @@ library LibUsdOracle {
             }
         }
 
-        // If the oracle implementation address is not set, use the current contract.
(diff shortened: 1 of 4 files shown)


Brean0:
Thinking about this more, Perhaps we should change LibUsdOracle to return 0, if the target address is not specified (and verify that the target address is non-zero upon whitelisting), given that Beanstalk does not have (and can't really have) a generic oracle implementation.

PR #1064, opened 2024-09-04
Fix: Incorrect Plot issuance in case `activeField` is updated before user's migration

nickkatsios:
Fixes [M-02]: Usually there is no problem because activeField is 0 (old version). However if user waited so long such that activeField was updated, migration would corrupt storage.


Brean0:
lgtm

PR #1067, opened 2024-09-05
Fix: Reseed Whitelist

nickkatsios:
This pr  addresses the `BeanstalkPrice` contract failing along with whitelist data errors in the reseed. 
- Update `ReseedWhitelist.sol` to set the oracle implementation for the non bean token in the well instead of the well lp token.
- Fix oracle settings in the whitelisted tokens json and add mocks.
- Perform a small swap in addLiquidity and transfer to avoid pump `NotInitialized()` error.
- Extract the newly added evaluation parameters and append them in the globals json.


Brean0:
lgtm!

protocol/reseed/reseedAddLiquidityAndTransfer.js
```diff
@@ -59,15 +58,17 @@ async function reseedAddLiquidityAndTransfer(account, L2Beanstalk, mock = true,
...
     await token.connect(account).approve(well.address, MAX_UINT256);
     await bean.connect(account).approve(well.address, MAX_UINT256);
     // add liquidity to well, to L2 Beanstalk:
-    console.log(`Adding liquidity to ${WellAddresses[i]}`);
+    console.log(`Adding liquidity to ${WellAddresses[i]} and performing a swap to update the well pump.`);
     await well
       .connect(account)
       .addLiquidity([beanAmounts[i], nonBeanAmounts[i]], 0, L2Beanstalk, MAX_UINT256);
+    // perform a swap to update the well pumps and avoid "NotInitialized" error.
+    await well.connect(account).swapFrom(bean.address, token.address, to6("1"), 0, account.address, MAX_UINT256);
```
Brean0, line 71: apologies, instead of swapping 1 bean here, I think we should call it with 0 beans instead (which should initialize the pump as well)

I would do this with `addLiquidity`(with 0)  as thats what was done in the past.

Brean0 (approved):
small comment, otherwise lgtm!

protocol/reseed/reseedAddLiquidityAndTransfer.js
nickkatsios: Fixed here and it worked: https://github.com/BeanstalkFarms/Beanstalk/pull/1067/commits/92a65bb3ed2aaff5ed8adf8e7b6e90184922ee59. Merging this now.

PR #1069, opened 2024-09-06
Fix: Set `excessFeeRefundAddress` to L2 receiver rather than L1 `msg.sender`

nickkatsios:
Fixes [M-03] from Timoh's review.
"In Arbitrum `excessFeeRefundAddress` is receiver of excess submission cost, worth noting is that excess fee is refunded on L2. https://docs.arbitrum.io/how-arbitrum-works/arbos/l1-l2-messaging#submission
Problem is that L2MigrationFacet.sol is explicitly assumed to be used by smart contracts which don't exist on L2. Therefore excess fee will be lost for them."


Brean0:
lgtm

PR #1073, opened 2024-09-09
Add External Unripe Data in Reseed

nickkatsios:
- Parses and adds data about external unripe bean and unripe LP holders.
- Deploys omitted L1ReceiverFacet from reseed 10


Brean0 (commented):
lgtm. Think that the external bean holder csv isn't used anywhere and can be deleted if needed.

PR #1075, opened 2024-09-10
`getWellTwaUsdLiquidityFromReserves` decimal fix. 

nickkatsios:
Addresses [L-05] from timoh and L-010 from codehawks:

When `getWellTwaUsdLiquidityFromReserves` is called by `SeasonGetterFacet` it calls `LibUsdOracle` which returns the non bean token price in a well with 6 decimal precision. before retunring the result, we divide by 1e6 so this precision is canceled out. We then multiuply by the twa reserves that have as many decimals as the `token` itself so the actual return value has as many decimals as `token`. The return value from `getWellTwaUsdLiquidityFromReserves` is then used in `LibEvaluate.calcLPToSupplyRatio` that assumes 18 decimal precision, so we need to account for whitelisted tokens that have less than 18 decimals by multiplying the precision difference.

### DIFF
--- a/protocol/contracts/libraries/Well/LibWell.sol
+++ b/protocol/contracts/libraries/Well/LibWell.sol
@@ -162,8 +162,15 @@ library LibWell {
         // if the function reaches here, then this is called outside the sunrise function
         // (i.e, seasonGetterFacet.getLiquidityToSupplyRatio()).We use LibUsdOracle
         // to get the price. This should never be reached during sunrise and thus
-        // should not impact gas.
-        return LibUsdOracle.getTokenPrice(token).mul(twaReserves[j]).div(1e6);
+        // should not impact gas. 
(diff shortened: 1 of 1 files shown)


Brean0:
Lgtm!

in `getWellTwaUsdLiquidityFromReserves`, `tokenUsd` is the value returned in `getUsdPrice`, which has the same decimal precision as the token being called. `twaReserves` also has the same decimal precision as the token. Thus:

1eN * 1e18 / 1eN = 1e18; 

However, in the case where tokenUsd = 1, we get: 

1e6 * 1eN / 1e6.

nickkatsios:
fixed here: https://github.com/BeanstalkFarms/Beanstalk/pull/1075/commits/f6cce8d40482f865118241bfa13bfb0ccfc7a273 and added the formula for clarity. Merging this now.

PR #1077, opened 2024-09-10
Fix: Set only individual receiver internal balance in `L1RecieverFacet`.

nickkatsios:
fixes [M-05] from timo.
- ReseedGlobal sets internal balances that also reflect smart contract account balances. 
- When performing the migration, all assets are optimistically minted to the diamond address. 
- When smart contracts migrate their internal balance via the `L2MigrationFacet` and `L1RecieverFacet`, an amount of assets will be allocated to their delegated `receiver`. This does not need to trigger an increase in global internal balances as its already accounted for in `ReseedGlobal`. 
- This way the effectiveness of the `fundsSafu` invariant modifier will be preserved.


Brean0:
lgtm

PR #1089, opened 2024-09-14
update libUsdOracle

Brean0:
- updates each whitelisted token to have a minimum stalkEarnedPerSeason of 1. This is not strictly necessary when whitelisting a token through the whitelistFacet, but `reseedWhitelist` skips the validation check that is performed. 
- Updates ReseedWhitelist such that the oracle implementation is only set on LP tokens.


nickkatsios:
LGTM

PR #1132, opened 2024-10-08
Final L2 deployment Tasks

nickkatsios:
- Remove ownership claim and facet addition from reseedL2.
- Add facet addition to generate the diamond cut to a new task `reseedL2AddFacets`.
- Fix 3crv pool usdc name.


Brean0:
lets roll

PR #1163, opened 2024-10-21
Update transmit in facet to accept bytes array for asset flexibility

pizzaman1337:
Updates TransmitInFacet to handle 2d bytes array for asset transmission flexibility.

(This PR will be updated to merge into the main transmission branch once the master merge branch is merged).

This also includes:
- Updates tests from using `require` to `assert`
- Adds test for `totalHarvested` and `totalSlashed`
- Disallows the transfer of germinating deposits
- Fixes TransmitOutFacet to pass through data from it's input param to the TransmitInFacet (bug fix)

### DIFF
--- a/protocol/contracts/beanstalk/ForkSystem/TransmitOutFacet.sol
+++ b/protocol/contracts/beanstalk/ForkSystem/TransmitOutFacet.sol
@@ -22,28 +22,31 @@ contract TransmitOutFacet is Invariable {
      */
     function transmitOut(
         address destination,
-        LibTransmitOut.SourceDeposit[] calldata sourceDeposits,
-        LibTransmitOut.SourcePlot[] calldata sourcePlots,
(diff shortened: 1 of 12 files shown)


protocol/contracts/beanstalk/ForkSystem/TransmitOutFacet.sol
```diff
@@ -25,7 +25,7 @@ contract TransmitOutFacet is Invariable {
         LibTransmitOut.SourceDeposit[] calldata sourceDeposits,
         LibTransmitOut.SourcePlot[] calldata sourcePlots,
         LibTransmitOut.SourceFertilizer[] calldata sourceFertilizer,
-        bytes calldata // data
+        bytes calldata data
```
pizzaman1337, line 28: @funderbrker I removed the `//` here so that the data var can be passed into the `transmitIn` function (line 47).

Brean0:
lgtm!

